PR #45, opened 2018-10-12
blob: add and expose Blob type

steveeJ:


src/v2/mod.rs
```diff
@@ -53,7 +53,8 @@ pub mod manifest;
 mod tags;
 pub use self::tags::StreamTags;
 
-mod blobs;
+pub mod blobs;
```
lucab: This exposes the whole module in the public API. It is not required here, and in general it is not something you want to do as it is easily results in exposing unintended internal details.

src/v2/mod.rs
```diff
@@ -53,7 +53,8 @@ pub mod manifest;
...
+pub use self::blobs::Blob;
```
lucab: You can merge this and the one below in a single `use` line.

src/v2/blobs.rs
```diff
@@ -3,8 +3,11 @@ use futures::Stream;
 use hyper::StatusCode;
 use v2::*;
 
+/// Convenience alias for the binary blob.
+pub type Blob = Vec<u8>;
```
lucab: What do you plan to use this for? I'm mildly against introducing aliases for plain stdlib types, as it increases the cognitive burden on developers consuming the library.

steveeJ:
Thanks for the input. This seems not worth the risks of exposing the whole module, I'm fine closing it.

lucab:
I'd say that in the future:
 * we may consider if we prefer an opaque newtype for blob (at the moment I don't see the need)
 * if we need to expose such a type, we can do selectively only for that (i.e. a `pub use` instead of a `pub mod`)

PR #47, opened 2018-10-14
v2/manifest: document layers order

lucab:
This documents the order used by this crate when listing
layers' digests.

### DIFF
--- a/src/v2/manifest/manifest_schema1.rs
+++ b/src/v2/manifest/manifest_schema1.rs
@@ -1,6 +1,8 @@
 use v2::*;
 
 /// Manifest version 2 schema 1, signed.
+///
+/// Specification is at https://docs.docker.com/registry/spec/manifest-v2-1/.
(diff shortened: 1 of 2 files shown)


steveeJ:
> he order is pretty much arbitrary, and docker switched it between two revisions, so I just picked the most recent one

If that's true we could detect the order and ensure we always hand it out in the same order. I don't have a strong opinion on what the order is, but I insist on it being predictable.

lucab:
@steveeJ yes, it's fixed to follow the v2s2 one (base layer first). For the old schema the inversion is done internally, here: https://github.com/camallo/dkregistry-rs/blob/2cd5f12ef1beb93006858bed06d07ea755f868f3/src/v2/manifest/manifest_schema1.rs#L43

steveeJ:
> yes, it's fixed to follow the v2s2 one (base layer first). For the old schema the inversion is done internally, here:

LGTM

PR #48, opened 2018-10-16
v2/tags: follow results on multiple pages

lucab:
This adds full support for tags pagination, with per-page limit
and next page following.

### DIFF
--- a/src/v2/tags.rs
+++ b/src/v2/tags.rs
@@ -5,65 +5,117 @@ use v2::*;
 /// Convenience alias for a stream of `String` tags.
 pub type StreamTags = Box<futures::Stream<Item = String, Error = Error>>;
 
+/// A chunk of tags for an image.
+///
+/// This contains a non-strict subset of the whole list of tags
+/// for an image, depending on pagination option at request time.
 #[derive(Debug, Default, Deserialize, Serialize)]
-struct Tags {
+struct TagsChunk {
+    /// Image repository name.
     name: String,
+    /// Subset of tags.
     tags: Vec<String>,
 }
 
 impl Client {
     /// List existing tags for an image.
     pub fn get_tags(&self, name: &str, paginate: Option<u32>) -> StreamTags {
-        let url = {
-            let mut s = format!("{}/v2/{}/tags/list", self.base_url, name);
-            if let Some(n) = paginate {
-                s = s + &format!("?n={}", n);
+        let dclient = self.clone();
+        let base_url = format!("{}/v2/{}/tags/list", self.base_url, name);
+
+        let fres = futures::stream::unfold(Some(String::new()), move |link| {
+            // Stream ends when response has no `Link` header.
+            let last = match link {
+                None => return None,
+                Some(ref s) if s == "" => None,
+                s => s,
             };
-            match hyper::Uri::from_str(s.as_str()) {
-                Ok(url) => url,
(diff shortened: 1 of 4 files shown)


lucab:
I've realised I've left a few unwraps around and forgot to update the mock tests. Still fine for a first review pass, but marked back as WIP.

steveeJ:
What are the missing steps in the mock tests? I can continue on top of your work, after #49, if you like

lucab:
@steveeJ this TODO https://github.com/camallo/dkregistry-rs/blob/86e85b4c70d54f81b9bd8722abe9f9a0c4777ffb/tests/mock/tags.rs#L81

lucab:
Got rid of the unwraps in the meanwhile.

lucab:
And the mock TODO too. This should be ready to review, but we may still want to land #49 first.

steveeJ:
This has conflicts due to the merge of #49. Let me know if you'd like me to pick it up.

lucab:
@steveeJ rebased to resolve conflicts.

lucab:
I've rebased this to current master. I guess we can tag 0.2.0 once this is merged.

src/v2/tags.rs
```diff
@@ -5,65 +5,116 @@ use v2::*;
...
-                    let ct_hdr = r.headers().get(header::CONTENT_TYPE);
+            let client = dclient.clone();
+            let url = hyper::Uri::from_str(&full_url);
+
+            let freq = futures::future::result(url)
+                .from_err()
+                .and_then(move |url| {
+                    trace!("GET {:?}", &url);
+                    let req = client.new_request(hyper::Method::GET, url);
+                    futures::future::result(req)
```
steveeJ: All these different future combinators look messy here. I've failed a quick try to reduce the nesting level down to one, but I'd really like to see that for readability.

src/v2/tags.rs
```diff
@@ -5,65 +5,116 @@ use v2::*;
...
+
         Box::new(fres)
     }
 }
+
+/// Parse a `Link` header.
+///
+/// Format is described at https://docs.docker.com/registry/spec/api/#listing-image-tags#pagination.
+fn parse_link(hdr: Option<header::HeaderValue>) -> Option<String> {
+    // TODO(lucab): find a better way to write this parser.
```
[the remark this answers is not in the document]
lucab: I've expanded the comment. It is a reminder to recheck for a library to do this (I didn't find any when I wrote it). I've left the author in the TODO as it common practice and makes it easier when code is copied/split/forked/moved around.

src/v2/tags.rs
lucab: This code is equivalent to https://github.com/camallo/dkregistry-rs/blob/23adf75204f96d8609ec1534bdefea63f678d9d8/src/v2/auth.rs#L27-L46

It seems quite concise/idiomatic, but I understand it's a subjective point.

lucab:
> I've already tagged 0.2.0

Oh. Then unfortunately your tag has a mismatch with the manifest version: https://github.com/camallo/dkregistry-rs/blob/0.2.0/Cargo.toml#L3. To avoid such problems in the future, I'll introduce [cargo-release](https://github.com/sunng87/cargo-release) metadata here too and use the same process we have for other coreos crate (which looks like https://github.com/coreos/openssh-keys/pull/31).

PR #49, opened 2018-10-17
API: substitute `Result<Future..>` with `Future..` for more idiomatic future chaining

steveeJ:
This change is motivated by an out-of-band discussion with @lucab and a comment on a related PR https://github.com/openshift/cincinnati/pull/4#discussion_r225639618 by @crawford, suggesting to use future chaining.

Before this PR all methods return a `Result<Future..>` chaining them requires to `unwrap()` or slightly milder `unwrap_or_else()` for *every* additional chain element, which adds significant boilerplate for the consumer. This PR aims to aid with this by removing the outer `Result<Future..>` and return `Future..` for all async public methods.

I'd like to discuss this API breaking change with the intention to move forward with this or a similar approach for the whole crate. I've done a rewrite of the login example demonstrates the usage of the changed API.

---
- [x] Rewrite *all of the* examples to use the new API
- Rewrite examples to use chaining
  - [x] ~~examples/checkregistry.rs~~
  - [x] examples/image.rs
  - [x] examples/login.rs
  - [x] examples/tags.rs
  - [x] examples/trace.rs
- [x] Rewrite *all of the?* API

### DIFF
--- a/examples/image.rs
+++ b/examples/image.rs
@@ -1,15 +1,17 @@
+extern crate dirs;
 extern crate dkregistry;
 extern crate futures;
 extern crate serde_json;
 extern crate tokio_core;
 
 use dkregistry::{reference, render};
+use futures::prelude::*;
+use std::result::Result;
+use std::str::FromStr;
 use std::{boxed, env, error, fs, io};
 use tokio_core::reactor::Core;
 
-use std::str::FromStr;
-
-type Result<T> = std::result::Result<T, boxed::Box<error::Error>>;
+mod common;
 
 fn main() {
     let dkr_ref = match std::env::args().nth(1) {
@@ -22,7 +24,7 @@ fn main() {
 
     let mut user = None;
     let mut password = None;
-    let home = env::home_dir().unwrap_or("/root".into());
+    let home = dirs::home_dir().unwrap();
     let cfg = fs::File::open(home.join(".docker/config.json"));
     if let Ok(fp) = cfg {
         let creds = dkregistry::get_credentials(io::BufReader::new(fp), &registry);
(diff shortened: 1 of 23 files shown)


src/v2/auth.rs
```diff
@@ -71,14 +71,15 @@ impl Client {
...
+    pub fn login(&self, scopes: &[&str]) -> FutureTokenAuth {
         let subclient = self.hclient.clone();
         let creds = self.credentials.clone();
         let scope = scopes
             .iter()
             .fold("".to_string(), |acc, &s| acc + "&scope=" + s);
         let auth = self
-            .get_token_provider()?
+            .get_token_provider()
+            .unwrap()
```
steveeJ: # FIXME

src/v2/mod.rs
```diff
@@ -103,13 +103,11 @@ impl Client {
...
 
-    pub fn is_v2_supported(&self) -> Result<FutureBool> {
+    pub fn is_v2_supported(&self) -> FutureBool {
         let api_header = "Docker-Distribution-API-Version";
         let api_version = "registry/2.0";
 
-        let url = try!(hyper::Uri::from_str(
-            (self.base_url.clone() + "/v2/").as_str()
-        ));
+        let url = hyper::Uri::from_str((self.base_url.clone() + "/v2/").as_str()).unwrap();
```
steveeJ: # FIXME

lucab:
I would say that yes, we should try to drop the outer `Result<>` and convert the error case into an inner future result. The current API is my first attempt at a futures-based library, as you can see...

steveeJ:
I've changed all methods referenced by the examples. A first review pass would make sense now ;-)

examples/image.rs
```diff
@@ -56,82 +58,110 @@ fn run(dkr_ref: &reference::Reference, user: Option<String>, passwd: Option<Stri
...
+        .insecure_registry(false)
+        .username(user)
+        .password(passwd)
+        .build()?;
+
+    let futures = futures::future::ok::<_, dkregistry::errors::Error>(&mut client)
+        .and_then(|dclient| {
+            dclient.is_v2_supported().and_then(|v2_supported| {
+                if !v2_supported {
+                    return Err("API v2 not supported".into());
```
lucab: You can drop the explicit `return` here (or flatten the else branch).

examples/image.rs
```diff
@@ -56,82 +58,110 @@ fn run(dkr_ref: &reference::Reference, user: Option<String>, passwd: Option<Stri
...
+                } else {
+                    Ok(dclient)
+                }
+            })
+        }).and_then(|dclient| {
+            dclient.is_auth(None).and_then(|is_auth| {
+                if is_auth {
+                    return Err("no login performed, but already authenticated".into());
```
lucab: Same as above.

examples/image.rs
```diff
@@ -56,82 +58,110 @@ fn run(dkr_ref: &reference::Reference, user: Option<String>, passwd: Option<Stri
...
+                .and_then(move |manifest_body| {
+                    let layers = match manifest_kind {
+                        dkregistry::mediatypes::MediaTypes::ManifestV2S1Signed => {
+                            let m: dkregistry::v2::manifest::ManifestSchema1Signed =
+                        serde_json::from_slice(manifest_body.as_slice()).unwrap();
+                            m.get_layers()
+                        }
+                        dkregistry::mediatypes::MediaTypes::ManifestV2S2 => {
+                            let m: dkregistry::v2::manifest::ManifestSchema2 =
+                            serde_json::from_slice(manifest_body.as_slice()).unwrap();
```
lucab: There are two leftover unwraps here and in the match-arm above.

examples/image.rs
```diff
@@ -56,82 +58,110 @@ fn run(dkr_ref: &reference::Reference, user: Option<String>, passwd: Option<Stri
...
-            out.len()
-        );
-        blobs.push(out);
+    // FIXME: use async io
+    let path = &format!("{}:{}", &image, &version).replace("/", "_");
+    let path = std::path::Path::new(&path);
+    if path.exists() {
+        return Err(format!("path {:?} already exists, exiting", &path).into());
     }
+    std::fs::create_dir(&path).unwrap();
```
lucab: For reference: https://docs.rs/tokio/0.1.12/tokio/fs/fn.create_dir_all.html

We are still using the old `tokio_core` everywhere in this repo, so I think we can leave this FIXME in place and tackle that once we plainly switch to the new `tokio` crate.

examples/login.rs
```diff
@@ -1,9 +1,12 @@
 extern crate dkregistry;
+extern crate futures;
 extern crate tokio_core;
 
 use std::{boxed, error};
 use tokio_core::reactor::Core;
 
+use futures::future::Future;
```
lucab: Just use the prelude here too.

examples/tags.rs
```diff
@@ -38,36 +38,57 @@ fn main() {
...
+            dclient.is_auth(None).and_then(|is_auth| {
+                if is_auth {
+                    Err("no login performed, but already authenticated".into())
+                } else {
+                    Ok(dclient)
+                }
+            })
+        }).and_then(|dclient| {
+            dclient
+                .login(&[&format!("repository:{}:pull", image)])
```
lucab: I would extract the scope-formatting outside (before) the start of the chain, so that it's harder for people to miss (here and everywhere else too).

examples/tags.rs
```diff
@@ -38,36 +38,57 @@ fn main() {
...
+                .and_then(|token| {
+                    dclient
+                        .is_auth(Some(token.token()))
+                        .and_then(move |is_auth| {
+                            if !is_auth {
+                                Err("login failed".into())
+                            } else {
+                                println!("logged in!");
+                                Ok(dclient
+                                    .clone()
```
lucab: Is this clone really needed?

src/v2/blobs.rs
```diff
@@ -33,11 +33,11 @@ impl Client {
     }
 
     /// Retrieve blob.
-    pub fn get_blob(&self, name: &str, digest: &str) -> Result<FutureBlob> {
+    pub fn get_blob(&self, name: &str, digest: &str) -> FutureBlob {
         let cl = self.clone();
         let url = {
             let ep = format!("{}/v2/{}/blobs/{}", self.base_url.clone(), name, digest);
-            hyper::Uri::from_str(ep.as_str())?
+            hyper::Uri::from_str(ep.as_str()).unwrap()
```
lucab: Leftover unwrap (there are a few more below without a fixme note).

examples/tags.rs
steveeJ: Yes, because:

```
error[E0597]: `dclient` does not live long enough
  --> examples/tags.rs:70:21
   |
69 |                 .and_then(|token| {
   |                           ------- capture occurs here
70 |                     dclient.is_auth(Some(token.token())).and_then(|is_auth| {
   |                     ^^^^^^^ borrowed value does not live long enough
...
82 |         }).and_then(|tags_collect| tags_collect);
   |         - borrowed value only lives until here
...
91 | }
   | - borrowed value needs to live until here

error[E0597]: `token` does not live long enough
  --> examples/tags.rs:76:49
   |
70 |                     dclient.is_auth(Some(token.token())).and_then(|is_auth| {
   |                                                                   --------- capture occurs here
...
76 |                                 .set_token(Some(token.token()))
   |                                                 ^^^^^ borrowed value does not live long enough
...
81 |                 })
   |                 - borrowed value only lives until here
...
91 | }
   | - borrowed value needs to live until here

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0597`.
error: Could not compile `dkregistry`.
```

examples/image.rs
steveeJ: I added a TODO here

src/v2/manifest/mod.rs
```diff
@@ -18,18 +18,20 @@ impl Client {
...
-    pub fn get_manifest(&self, name: &str, reference: &str) -> Result<FutureManifest> {
-        let url = try!(hyper::Uri::from_str(&format!(
+    pub fn get_manifest(&self, name: &str, reference: &str) -> FutureManifest {
+        let url = hyper::Uri::from_str(&format!(
             "{}/v2/{}/manifests/{}",
             self.base_url.clone(),
             name,
             reference
-        )));
+        )).unwrap();
```
lucab: Leftover unwrap.

src/v2/manifest/mod.rs
```diff
@@ -18,18 +18,20 @@ impl Client {
...
         let req = {
             let mut r = self.new_request(hyper::Method::GET, url.clone());
             let mtype = mediatypes::MediaTypes::ManifestV2S2.to_string();
-            r.headers_mut()
-                .append(header::ACCEPT, header::HeaderValue::from_str(&mtype)?);
+            r.headers_mut().append(
+                header::ACCEPT,
+                header::HeaderValue::from_str(&mtype).unwrap(),
```
lucab: Leftover unwrap.

src/v2/manifest/mod.rs
```diff
@@ -61,19 +63,19 @@ impl Client {
...
+    ) -> mediatypes::FutureMediaType {
         let url = {
             let ep = format!(
                 "{}/v2/{}/manifests/{}",
                 self.base_url.clone(),
                 name,
                 reference
             );
-            try!(hyper::Uri::from_str(ep.as_str()))
+            hyper::Uri::from_str(ep.as_str()).unwrap()
```
lucab: Leftover unwrap.

src/v2/manifest/mod.rs
```diff
@@ -61,19 +63,19 @@ impl Client {
...
         };
         let accept_types = match mediatypes {
             None => vec![mediatypes::MediaTypes::ManifestV2S2.to_mime()],
-            Some(ref v) => try!(to_mimes(v)),
+            Some(ref v) => to_mimes(v).unwrap(),
```
lucab: Leftover unwrap.

src/v2/tags.rs
```diff
@@ -13,13 +13,13 @@ struct Tags {
...
     /// List existing tags for an image.
-    pub fn get_tags(&self, name: &str, paginate: Option<u32>) -> Result<StreamTags> {
+    pub fn get_tags(&self, name: &str, paginate: Option<u32>) -> StreamTags {
         let url = {
             let mut s = format!("{}/v2/{}/tags/list", self.base_url, name);
             if let Some(n) = paginate {
                 s = s + &format!("?n={}", n);
             };
-            try!(hyper::Uri::from_str(s.as_str()))
+            hyper::Uri::from_str(s.as_str()).unwrap()
```
lucab: Leftover unwrap.

lucab:
@steveeJ this looks pretty good to me. There are only five leftover unwraps and then I think it's done.

steveeJ:
Yay, more unwraps to clear up :birthday: Thanks for catching these

PR #52, opened 2018-11-01
docs: add release guide and metadata

lucab:


docs/release-guide.md
```diff
@@ -0,0 +1,78 @@
...
+#### 4. Get the PR reviewed, approved and merged
+
+#### 5. Publish the artifacts (tag and crate)
+
+* `git push --tags ${UPSTREAM_REMOTE} ${RELEASE_VER}`
+* Make sure the upstream tag matches the local tag:
+
+    `git fetch --tags --verbose ${UPSTREAM_REMOTE} 2>&1 | grep ${RELEASE_VER}`
+* `git checkout ${RELEASE_VER}`
+* Make sure the tag is what you released and you get; this will show an empty output:
```
steveeJ: There's a cut in the first part. Better:

Make sure the tag is what you intend to release; if so this will show an empty output:

docs/release-guide.md
```diff
@@ -0,0 +1,78 @@
...
+
+    `git diff release-${RELEASE_VER}`
+* `cargo publish`
```
steveeJ: Do we really want to make this an unconditional part of the process?

lucab: Yes, this is the main artifact of the whole process and how Rust libraries are consumed. If a crate should not be pushed to the registry, it should have `publish = false` in its manifest to prevent that.

docs/release-guide.md
lucab: Oops, I failed at this. Updated with your wording now.

Issue #44, opened 2018-10-12
v2/client: move to async reqwest


lucab:
Now that reqwest has an [async](https://docs.rs/reqwest/0.9.2/reqwest/async/index.html) module, we should use that instead of plain hyper. Unfortunately, this means switching from rustls to native-tls, thus getting a FFI dependency to openssl (or similar).

---
Progress
- [x] src/v2/blobs.rs
- [x] src/v2/tags.rs
- [x] src/v2/manifest/mod.rs
- [ ] src/v2/auth.rs #95 
- [ ] src/v2/catalog.rs #95
- [ ] src/v2/mod.rs #95
- [ ] src/v2/config.rs #95 
- [ ] src/lib.rs #95
- [ ] src/errors.rs #95

lucab:
@steveeJ both are still in development and will likely see a major rework once [futures land in stdlib](https://doc.rust-lang.org/nightly/std/future/index.html). However hyper is the wrong layer to use when implementing an HTTP client, which means that we poorly re-implement here some middleware logic (e.g. [following redirections](https://github.com/camallo/dkregistry-rs/blob/bacca3b7fb80edf7c38035d6f355672aa2536dd4/src/v2/blobs.rs#L48-L83)).

This crate status quo is is purely a result of historical timing, as at that point there was no higher-level async HTTP client.

steveeJ:
> However hyper is the wrong layer to use when implementing an HTTP client, which means that we poorly re-implement here some middleware logic (e.g. [following redirections](https://github.com/camallo/dkregistry-rs/blob/bacca3b7fb80edf7c38035d6f355672aa2536dd4/src/v2/blobs.rs#L48-L83)).
> 
> This crate status quo is is purely a result of historical timing, as at that point there was no higher-level async HTTP client.

I understand your concern. If we need to extend the functionality of dkregistry we can refactor the code to use *reqwest* along that. Personally I don't have the urge to do it earlier, but of course I'm not stopping you from doing it ;-)

steveeJ:
The close was unintentional via a stale `Fixes ..` in a PR.
Marking this as a bug because it causes the system certificate store to be ignored, which is the default behavior of `hyper-rustls`.

PR #63, opened 2018-11-14
blobs: return an error on unhandled http status code

steveeJ:
Before this change even error codes would be treated as success and the
body would contain the HTML formatted error message passed through to
the caller.

### DIFF
--- a/src/v2/blobs.rs
+++ b/src/v2/blobs.rs
@@ -82,7 +82,14 @@ impl Client {
                     | StatusCode::FOUND => {
                         trace!("Got moved status {:?}", r.status());
                     }
-                    _ => return Either::A(future::ok(r)),
+                    StatusCode::OK => {
+                        return Either::A(future::ok(r));
+                    }
+                    _ => {
+                        return Either::A(future::err(
+                            format!("unexpected status code '{}'", r.status()).into(),
+                        ));
+                    }
                 };
                 let redirect: Option<String> = match r.headers().get("Location") {
                     None => {


src/v2/blobs.rs
```diff
@@ -82,7 +82,14 @@ impl Client {
                     | StatusCode::FOUND => {
                         trace!("Got moved status {:?}", r.status());
                     }
-                    _ => return Either::A(future::ok(r)),
+                    StatusCode::OK => {
```
lucab: This closure is only meant to handle top-level redirections. This whole closure at some point should be replaced by https://github.com/camallo/dkregistry-rs/issues/44.

The error status check is correct, but I suggest not to place it here. A better place would be a dedicated `and_then` after this one and before body concatenation (or directly in that closure).

steveeJ: Looking at it more closely it would be necessary to construct a loop which handles a chain of requests until it falls into a case where the final state is known. E.g. you could have multiple redirections which would slip through here too.

lucab: Indeed. That's the point of https://github.com/camallo/dkregistry-rs/issues/44#issuecomment-430135569.

steveeJ: I see. I'll give it a shot to use *reqwest* for the *blobs* module.

lucab: I would be happy to see that, but I don't know if it is possible without reworking the whole client. If your experiment starts leaning toward to the latter, I think you'd get better value by landing just the status-checking in this PR and split the reqwest rework.

PR #65, opened 2018-11-15
blobs: switch from hyper to reqwest

steveeJ:
This change is motivated by improved status error handling.
Before this change even error codes would be treated as success and the
body would contain the HTML formatted error message passed through to
the caller.
Instead of reimplementing proper error handling here we can make use of
the reqwest library which already does it.

---

Closes #63 
Progress on #44

### DIFF
--- a/src/v2/blobs.rs
+++ b/src/v2/blobs.rs
@@ -1,6 +1,6 @@
-use futures::future::{self, Either};
 use futures::Stream;
-use hyper::StatusCode;
+use reqwest;
+use reqwest::StatusCode;
 use v2::*;
 
 /// Convenience alias for future binary blob.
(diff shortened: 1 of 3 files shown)


src/v2/blobs.rs
```diff
@@ -62,60 +47,27 @@ impl Client {
...
+            .inspect(|res| trace!("Blob GET result: {:?}", res.status()))
+            .and_then(|mut res| {
+                let body = std::mem::replace(res.body_mut(), reqwest::async::Decoder::empty());
+                body.concat2()
+            }).map_err(|e| Error::from(format!("{}", e)))
+            .and_then(|body| {
+                let mut cursor = std::io::Cursor::new(body);
+                let mut buf = Vec::new();
+                use std::io::Read;
+                match cursor.read_to_end(&mut buf) {
```
lucab: This is a blocking operation which would be better not to perform inside a future chain. I'm not sure what you are trying to do here, transforming a `Chunk` a into `Vec<u8>`?

src/v2/blobs.rs
```diff
@@ -62,60 +47,27 @@ impl Client {
...
-                        )))
-                    }
-                    Some(loc) => {
-                        trace!("Got Location header {:?}", loc);
-                        String::from_utf8(loc.as_bytes().to_vec()).ok()
+
+        let fres = reqwest::async::Client::new()
+            .get(url)
+            .send()
+            .inspect(|res| trace!("Blob GET result: {:?}", res.status()))
```
lucab: If this PR replaces the other one, then you probably want to check response status at this point, right?

lucab:
@steveeJ does this close #63 or is it a prerequisite for that? I didn't see the HTTP status check here.

src/v2/blobs.rs
steveeJ: >  I'm not sure what you are trying to do here, transforming a `Chunk` into `Vec<u8>`

Yes.

> This is a blocking operation which would be better not to perform inside a future chain.
 
I'd be happy to receive some help with one then :-)

steveeJ:
> does this close #63 or is it a prerequisite for that? I didn't see the HTTP status check here.

TL;DR: It does now, thanks for pointing this out :smile: 

I thought it already did because I was able to trigger an error when I cross tested the changes from cincinnati, but I must've made a mistake at one point.

src/v2/blobs.rs
```diff
@@ -62,60 +47,32 @@ impl Client {
...
-                        trace!("Got Location header {:?}", loc);
-                        String::from_utf8(loc.as_bytes().to_vec()).ok()
+
+        let fres = reqwest::async::Client::new()
+            .get(url)
+            .send()
+            .inspect(|res| trace!("Blob GET status: {:?}", res.status()))
+            .and_then(|mut res| {
+                let body =
+                    std::mem::replace(res.body_mut(), reqwest::async::Decoder::empty()).concat2();
```
lucab: This looks a bit funky, but I'm not familiar enough with it to know if there is a more straightforward way to do this.

src/v2/blobs.rs
```diff
@@ -62,60 +47,32 @@ impl Client {
...
-                    return Either::B(cl.hclient.request(req).from_err());
-                };
-                Either::A(future::ok(r))
-            }).and_then(|r| {
-                r.into_body()
-                    .concat2()
-                    .map_err(|e| format!("get_blob: failed to fetch the whole body: {}", e).into())
-            }).and_then(|body| Ok(body.into_bytes().to_vec()));
+                    status => Err(Error::from(format!(
+                        "GET request failed with status '{}' and body of size {}: {:#?}",
```
lucab: Minor nit: I think that this kind of status check is usually performed before starting to collect the whole body (to skip useless work) and the return error doesn't usually carry around the whole response (which could potentially be quite huge). I don't know if this can apply here and I'll leave the ergonomics to you, I just wanted to mention that.

steveeJ: https://www.freesoft.org/CIE/RFC/2068/89.htm contains that on 4xx the body should be displayed to the user. I'll give filtering this a shot.

lucab: Ack. I'm fine both ways. Leaving this to your discretion.

src/v2/blobs.rs
steveeJ: I've tried to avoid the `std::mem::replace` but couldn't find a way to satisfy the borrow checker, The lifetime of `res` is not long enough, and it also can't be cloned. This seems to leave no other safe option as using `replace` here.
I found a document which explains it in more detail: https://github.com/rust-unofficial/patterns/blob/master/idioms/mem-replace.md

PR #68, opened 2018-11-15
Reuse the token set by hyper for reqwest calls

### LINKED ISSUE #66: blobs: authentication is broken
Since the token is appended to the request headers of the *hyper* client, which is not considered *reqwest*, #65 broke authenticated calls for blobs.

steveeJ:
Depends on #69.
Fixes #66

### DIFF
--- a/src/v2/mod.rs
+++ b/src/v2/mod.rs
@@ -140,6 +140,18 @@ impl Client {
             });
         Box::new(fres)
     }
+
+    /// Takes reqwest's async RequestBuilder and injects an authentication header if a token is present
+    fn build_reqwest(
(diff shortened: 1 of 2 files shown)


src/v2/mod.rs
```diff
@@ -142,6 +142,21 @@ impl Client {
     }
 }
 
+/// Reqwest related implementation
+impl Client {
+    /// Takes reqwest's async RequestBuilder and injects an authentication header if a token is present
+    fn build_reqwest(
```
lucab: Let's have this in the already existing `impl Client` block above, as it will eventually be moved there anyway.

PR #70, opened 2018-11-15
tests/quay: test authenticated methods

lucab:
Add a login test for quay.io. Credentials for a dedicated CI user
are set up on travis.

### DIFF
--- a/tests/mod.rs
+++ b/tests/mod.rs
@@ -1,5 +1,9 @@
 #[cfg(feature = "test-net")]
 mod net;
 
+#[cfg(feature = "test-net")]
+#[macro_use]
+extern crate error_chain;
+
 #[cfg(feature = "test-mock")]
 mod mock;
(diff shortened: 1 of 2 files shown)


lucab:
Uhm, this worked on `master` but this PR is not seeing user/passwd vars.

@steveeJ it's a free "dkregistry" user on quay/dockerhub. It doesn't have any repo but we can grant read access to existing private ones.

steveeJ:
> Uhm, this worked on `master` but this PR is not seeing user/passwd vars.
Probably some deeply hidden travis setting.

> it's a free "dkregistry" user on quay/dockerhub. It doesn't have any repo but we can grant read access to existing private ones.

I granted that user read access to *steveej/cincinnati-test* which will be available for the time being.

lucab:
@steveeJ changed to use your pull-secrets. This covers login+tags+manifest right now. We can land it first and add more things later, or I can add a few more here, as you prefer.

lucab:
This now also covers fetching the blob from a single-layer private image, PTAL.

PR #76, opened 2018-12-12
dkregistry: remove tokio-core, use tokio instead

steveeJ:
@lucab PTAL and let me know if I'm missing something or if the handle is indeed unused

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ serde_json = "1"
 strum = "0.13"
 strum_macros = "0.13"
 tar = "0.4"
-tokio-core = "0.1"
+tokio = "0.1"
 dirs = "1.0"
 reqwest = "0.9"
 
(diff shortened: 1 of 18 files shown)


lucab:
We can remove it, I think it was required by the previous hyper async API. Can we fully drop `tokio-core` from cargo manifest after this rework?

steveeJ:
> Can we fully drop `tokio-core` from cargo manifest after this rework?

After changing the occurrences in all examples etc I believe we can. Since you agree I'm going to proceed and we'll find out for sure.

lucab:
@steveeJ yes, please. And thanks a lot for the rework!

lucab:
@steveeJ any plan to finish this? I have some other cincinnati changes that may benefit from it.

lucab:
I pushed forward the changes in your second commit, using `tokio` runtime instead and completely dropping `tokio-core`. Travis is green and LGTM. Please have a last look and merge if ok.

steveeJ:
> I pushed forward the changes in your second commit, using `tokio` runtime instead and completely dropping `tokio-core`. Travis is green and LGTM. Please have a last look and merge if ok.

Nice, thanks for picking this one up! I had another look and couldn't spot any issues.

PR #82, opened 2019-01-12
mediatypes: add band-aid for quay.io handling

steveeJ:
Since recently quay.io appends `; charset=utf-8` in the `content-type`
header value for HEAD requests on the `manifests` endpoint. This patch
restores functionality as before the quay.io change.

### DIFF
--- a/src/mediatypes.rs
+++ b/src/mediatypes.rs
@@ -17,7 +17,12 @@ pub enum MediaTypes {
     #[strum(props(Sub = "vnd.docker.distribution.manifest.v1+json"))]
     ManifestV2S1,
     /// Signed manifest, version 2 schema 1.
-    #[strum(serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws")]
+    #[strum(
+        serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws",
+
+        // TODO(steveeJ) find a generic way to handle this form
+        serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws; charset=utf-8",
+    )]
     #[strum(props(Sub = "vnd.docker.distribution.manifest.v1+prettyjws"))]
     ManifestV2S1Signed,
     /// Manifest, version 2 schema 2.


src/mediatypes.rs
```diff
@@ -18,6 +18,9 @@ pub enum MediaTypes {
     ManifestV2S1,
     /// Signed manifest, version 2 schema 1.
     #[strum(serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws")]
+    #[strum(
+        serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws; charset=utf-8"
```
crawford: Minor style suggestion: I would add this to the above attribute:

```rust
#[strum(
    serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws",
    serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws; charset=utf-8"
)]
```

lucab:
This PR is unfortunately wrong (and I fear the change on quay.io side too).

Media-types are part of the docker-registry API specifications. The V2S1 types are defined [here](https://docs.docker.com/registry/spec/manifest-v2-1/) and they don't have any additional (including charset) parameters.

Additionally, those manifests are explicitly compatible with `application/json`. JSON is a [binary media type](https://www.iana.org/assignments/media-types/application/json). As binary, it doesn't really have an outer charset, and the spec explicitly says that it doesn't have any additional (including charset) parameters.

@thomasmckay @josephschorr any chance to revert the change on quay, and align it back with IETF and docker-registry specs?

josephschorr:
@lucab We can certainly revert, but we'd have to double check on what Docker is returning for manifests. The problem is the Docker documentation itself shows `Content-Type: application/json; charset=utf-8` on some of the API returns, so it isn't clear whether that applies to the manifest type when it itself is `application/json`. I want to be sure before we go back and potentially break utf-8 encoded manifests.

lucab:
For reference, the quay.io change was reverted in the meanwhile. It was breaking kubernetes CI too: https://github.com/kubernetes/kubernetes/issues/72863#issuecomment-453926310

PR #85, opened 2019-01-24
 v2/manifest: add get_manifest_and_ref method

steveeJ:
This method returns the manifestref in addition to the manifest.  As
part of the migration to reqwest the method is a rewrite of
`get_manifest(..)`, which now wraps the new one and omits the
manifestref from the result.

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -19,62 +19,81 @@ impl Client {
     /// The name and reference parameters identify the image.
     /// The reference may be either a tag or digest.
     pub fn get_manifest(&self, name: &str, reference: &str) -> FutureManifest {
-        let url = match hyper::Uri::from_str(&format!(
-            "{}/v2/{}/manifests/{}",
-            self.base_url.clone(),
-            name,
-            reference
-        )) {
-            Ok(url) => url,
+        Box::new(
+            self.get_manifest_and_ref(name, reference)
+                .map(|(manifest, _)| manifest),
+        )
+    }
+
+    /// Fetch an image manifest and return it with its digest.
+    ///
+    /// The name and reference parameters identify the image.
+    /// The reference may be either a tag or digest.
+    pub fn get_manifest_and_ref(&self, name: &str, reference: &str) -> FutureManifestAndRef {
+        let url = {
+            let ep = format!(
+                "{}/v2/{}/manifests/{}",
+                self.base_url.clone(),
+                name,
+                reference
+            );
+            match reqwest::Url::parse(&ep) {
+                Ok(url) => url,
+                Err(e) => {
+                    return Box::new(future::err::<_, _>(Error::from(format!(
+                        "failed to parse url from string '{}': {}",
+                        ep, e
+                    ))));
+                }
+            }
+        };
+
+        let mtype = match header::HeaderValue::from_str(
+            &mediatypes::MediaTypes::ManifestV2S2.to_string(),
+        ) {
+            Ok(headervalue) => headervalue,
             Err(e) => {
-                let msg = format!("failed to parse Uri from str: {}", e);
+                let msg = format!("failed to parse HeaderValue from str: {}:", e);
                 error!("{}", msg);
                 return Box::new(futures::future::err::<_, _>(Error::from(msg)));
             }
         };
-        let req = {
-            let mut req = match self.new_request(hyper::Method::GET, url.clone()) {
-                Ok(r) => r,
-                Err(e) => {
-                    let msg = format!("new_request failed: {}", e);
-                    error!("{}", msg);
-                    return Box::new(futures::future::err(Error::from(msg)));
-                }
-            };
-            let mtype = mediatypes::MediaTypes::ManifestV2S2.to_string();
-            req.headers_mut().append(
-                header::ACCEPT,
-                match header::HeaderValue::from_str(&mtype) {
-                    Ok(headervalue) => headervalue,
-                    Err(e) => {
-                        let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                        error!("{}", msg);
-                        return Box::new(futures::future::err::<_, _>(Error::from(msg)));
-                    }
-                },
-            );
-            req
-        };
-        let freq = self.hclient.request(req);
-        let fres = freq
-            .from_err()
-            .inspect(move |_| {
-                trace!("GET {:?}", url);
-            })
-            .and_then(|r| {
-                let status = r.status();
+
+        let fres = self
+            .build_reqwest(
+                reqwest::async::Client::new()
+                    .get(url)
+                    .header(reqwest::header::ACCEPT, mtype),
+            )
+            .send()
+            .map_err(|e| Error::from(format!("{}", e)))
+            .and_then(|res| {
+                trace!("GET '{}' status: {:?}", res.url(), res.status());
+                let status = res.status();
                 trace!("Got status: {:?}", status);
                 match status {
-                    hyper::StatusCode::OK => Ok(r),
-                    _ => Err(format!("get_manifest: wrong HTTP status '{}'", status).into()),
+                    reqwest::StatusCode::OK => Ok(res),
+                    _ => Err(format!("GET {}: wrong HTTP status '{}'", res.url(), status).into()),
                 }
             })
-            .and_then(|r| {
-                r.into_body().concat2().map_err(|e| {
-                    format!("get_manifest: failed to fetch the whole body: {}", e).into()
-                })
+            .and_then(|res| {
+                future::ok(res.headers().clone()).join(
+                    res.into_body()
+                        .concat2()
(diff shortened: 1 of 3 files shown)


lucab:
@steveeJ a couple of questions:
 * do we want to introduce a structure holding both manifest-blob and ref?
 * do we want to directly replace current `get_manifest` with this?

steveeJ:
> * do we want to introduce a structure holding both manifest-blob and ref?

I'd say yes, because they are both retrieved in the same API call. I don't see anything wrong with getting both at once.

> * do we want to directly replace current `get_manifest` with this?
 
They're almost equivalent, except that one returns the manifestref too, which is simply omitted. I trust Rust to be sufficiently efficient here. Since the existing tests aren't failing it at least is shown to work :-)

steveeJ:
I want to explain the context a bit. In cincinnati we need the manifestref in order to get the labels from the separate quay API. Passing the manifestref on like this PR implements it is the most efficient way obvious to me. See [the relevant code in a snapshot of my working branch](https://github.com/steveeJ-forks/openshift-cincinnati/blob/add-quay-labels-to-graph/graph-builder/src/registry.rs#L121-L141).

lucab:
@steveeJ I think both points above makes sense to implement. Do you prefer to land this as-is or to make the changes now?

steveeJ:
> I think both points above makes sense to implement. Do you prefer to land this as-is or to make the changes now?

I'd say it can be merged as is now.
To clarify I interpreted your questions differently earlier. Let me go again:

> do we want to introduce a structure holding both manifest-blob and ref?

I think that's not necessary since we have the type alias which is convenient enough.

> do we want to directly replace current `get_manifest` with this?

Not necessarily since it would break compatibility with the existing tests and other code there might be.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
-            .inspect(move |_| {
-                trace!("GET {:?}", url);
-            })
-            .and_then(|r| {
-                let status = r.status();
+
+        let fres = self
+            .build_reqwest(reqwest::async::Client::new().get(url))
+            .send()
+            .map_err(|e| ::errors::Error::from(format!("{}", e)))
```
lucab: I think `Error` is already in scope here and can be used directly.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
                 }
             })
-            .and_then(|r| {
-                r.into_body().concat2().map_err(|e| {
-                    format!("get_manifest: failed to fetch the whole body: {}", e).into()
-                })
+            .and_then(|mut res| {
+                std::mem::replace(res.body_mut(), reqwest::async::Decoder::empty())
+                    .concat2()
+                    .map_err(|e| ::errors::Error::from(format!("{}", e)))
```
lucab: I think `Error` is already in scope here and can be used directly.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
+                    .join(futures::future::ok(res))
             })
-            .and_then(|body| Ok(body.into_bytes().to_vec()));
+            .map_err(|e| ::errors::Error::from(format!("{}", e)))
```
lucab: I think `Error` is already in scope here and can be used directly.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
-                    return Box::new(futures::future::err(Error::from(msg)));
+                    return Box::new(futures::future::err::<_, _>(Error::from(format!(
+                        "failed to parse url from string '{}': {}",
+                        ep, e
+                    ))));
                 }
-            };
-            let mtype = mediatypes::MediaTypes::ManifestV2S2.to_string();
-            req.headers_mut().append(
-                header::ACCEPT,
```
lucab: Did you drop this logic on purpose?

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
+                    .concat2()
+                    .map_err(|e| ::errors::Error::from(format!("{}", e)))
+                    .join(futures::future::ok(res))
             })
-            .and_then(|body| Ok(body.into_bytes().to_vec()));
+            .map_err(|e| ::errors::Error::from(format!("{}", e)))
+            .and_then(|(body, res)| {
+                let body_vec = body.to_vec();
+                let len = body_vec.len();
+                let status = res.status();
```
lucab: I think both `len` and `status` are unused.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
+                    reqwest::StatusCode::OK => Ok(res),
+                    _ => Err(format!("GET {}: wrong HTTP status '{}'", res.url(), status).into()),
                 }
             })
-            .and_then(|r| {
-                r.into_body().concat2().map_err(|e| {
-                    format!("get_manifest: failed to fetch the whole body: {}", e).into()
-                })
+            .and_then(|mut res| {
+                std::mem::replace(res.body_mut(), reqwest::async::Decoder::empty())
```
lucab: Is this equivalent to `into_body` or is there a reason to prefer this form? I personally find it a bit obscure.

src/v2/manifest/mod.rs
```diff
@@ -19,62 +19,68 @@ impl Client {
...
-            .and_then(|body| Ok(body.into_bytes().to_vec()));
+            .map_err(|e| ::errors::Error::from(format!("{}", e)))
+            .and_then(|(body, res)| {
+                let body_vec = body.to_vec();
+                let len = body_vec.len();
+                let status = res.status();
+                Ok((
+                    body_vec,
+                    res.headers()
+                        .get("docker-content-digest")
```
lucab: This seems to be with uppercase initials in registry API specs: https://docs.docker.com/registry/spec/api/#digest-header

steveeJ: quay doesn't follow that apparently:

```console
curl -v --get https://quay.io/v2/coreos/etcd/manifests/latest
(...)
* Connection state changed (MAX_CONCURRENT_STREAMS == 128)!
< HTTP/2 200 
< server: nginx/1.14.2
< date: Thu, 24 Jan 2019 17:49:22 GMT
< content-type: application/vnd.docker.distribution.manifest.v1+prettyjws
< content-length: 6655
< docker-content-digest: sha256:5b6691b7225a3f77a5a919a81261bbfb31283804418e187f7116a0a9ef65d21d
< x-frame-options: DENY
< strict-transport-security: max-age=63072000; preload
(...) 
```

lucab: Ah, they are case-insensitive according to https://tools.ietf.org/html/rfc7230#section-3.2. It would be nice to use the spec format if the getter supports this internally. Otherwise don't mind.

steveeJ: I verified that the current code is case-insensitive

src/v2/manifest/mod.rs
steveeJ: No, though it doesn't seem required. Should I add it back?

src/v2/manifest/mod.rs
steveeJ: It seems to be the only way to take ownership of the body. I've tried to use `into_body()` without success, as passing `ref` leads to an error:

```
error[E0382]: use of moved value: `res`
  --> src/v2/manifest/mod.rs:84:38
   |
81 |                 res.into_body()
   |                 --- value moved here
...
84 |                     .join(future::ok(res))
   |                                      ^^^ value used here after move
```

src/v2/manifest/mod.rs
steveeJ: Still unsure but I added it back to keep the semantics unchanged

lucab: This is a "SHOULD" according to registry spec. I don't remember exactly, but I think docker hub may require this.

src/v2/manifest/mod.rs
lucab: Eh, the compiler is always right. I think that may work if you reverse the order of the two logical operations in the chain, ie. first extracting the header value and then collecting the body.

steveeJ: I found the following to work too, would you prefer that?

```rust
            .and_then(|res| {
                future::ok(res.headers().clone()).join(
                    res.into_body()
                        .concat2()
                        .map_err(|e| Error::from(format!("{}", e))),
                )
            })
```

src/v2/manifest/mod.rs
lucab: @steveeJ you didn't change the header name constant though. Was that on purpose or you just forgot to?

steveeJ: I left it lower-case cause I thought that makes more sense from a perspective of being case-insensitive. At least that's the config I'm using in my code editors: case-insensitive if all-lowercase.

PR #86, opened 2019-01-24
Don't pull in openssl.

khuey:
I'm not sure if you actually need TLS for reqwest but if you do lets not pull in openssl since you're already using rustls.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ strum_macros = "0.13"
 tar = "0.4"
 tokio = "0.1"
 dirs = "1.0"
-reqwest = "0.9"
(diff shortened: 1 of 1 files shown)


lucab:
@khuey thanks, it makes sense. I think longer term we are going to use reqwest only (https://github.com/camallo/dkregistry-rs/issues/44), at which point we should just re-expose its TLS-related feature knobs to consumers.

lucab:
@khuey heads in case you are using this somewhere. We completed the switch to reqwest and removed direct usage of hyper-rustls. https://github.com/camallo/dkregistry-rs/pull/96 allows to switch the underlying TLS library, reaching the same goal of this PR.

khuey:
Thanks!

Issue #89, opened 2019-02-01
v2/client: friction around insecure registries and the `is_v2_supported` method


steveeJ:
IMHO the result of the `is_v2_supported()` method isn't correct as it is. 
In case of querying a v2-supported registry via plain http, the 301 status code (MOVED) will cause the method to return `false` to indicate that v2 is not supported here. 
The correct behavior would be look at the new URL and if the scheme changed to https, instead of returning false, throw an error which reports that the requested scheme is not supported.

I'll go ahead and call this a bug, but I'd like to know the original motivation behind this in case I'm missing something.

lucab:
The original motivation is that specs do not contemplate redirection there: https://docs.docker.com/registry/spec/api/#api-version-check

However I do agree that for a client to be helpful it should be able to handle redirections on most of the paths. Also, all common registries seems to have a redirection in place for HTTP to HTTPS.

PR #90, opened 2019-02-04
v2: switch version probing to reqwest

lucab:
This changes the v2 probing logic to use the reqwest client, in order
to better support redirection.

Closes: https://github.com/camallo/dkregistry-rs/issues/89

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -3,6 +3,7 @@
 use base64;
 use http;
 use hyper;
+use reqwest;
 use serde_json;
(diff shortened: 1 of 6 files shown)


lucab:
@steveeJ I was on a plane when writing this, so didn't realize you had https://github.com/camallo/dkregistry-rs/pull/91 in parallel. Feel free to take the bits you can re-use from here, and then close this PR. I'll review #91 when ready, no hurry, and sorry for the duplication.

PR #91, opened 2019-02-04
v2: partially substitute hyper for reqwest and work around Satellite registry

### LINKED ISSUE #44: v2/client: move to async reqwest

### LINKED ISSUE #89: v2/client: friction around insecure registries and the `is_v2_supported` method

steveeJ:
Fixes #44 
Partially fixes #89.
Closes #90.

### DIFF
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -25,32 +25,27 @@ type FutureString = Box<futures::Future<Item = String, Error = self::Error>>;
 impl Client {
     fn get_token_provider(&self) -> FutureString {
         let url = {
-            let ep = format!("{}/v2/", self.base_url);
-            match hyper::Uri::from_str(ep.as_str()) {
+            let ep = format!("{}/v2/", self.base_url.clone(),);
+            match reqwest::Url::parse(&ep) {
                 Ok(url) => url,
                 Err(e) => {
-                    return Box::new(futures::future::err::<_, _>(Error::from(format!(
-                        "failed to parse url from string: {}",
-                        e
+                    return Box::new(future::err::<_, _>(Error::from(format!(
+                        "failed to parse url from string '{}': {}",
+                        ep, e
                     ))));
                 }
             }
         };
-        let req = match self.new_request(hyper::Method::GET, url) {
-            Ok(r) => r,
-            Err(e) => {
-                let msg = format!("new_request failed: {}", e);
-                error!("{}", msg);
-                return Box::new(futures::future::err::<_, _>(Error::from(msg)));
-            }
-        };
-        let freq = self.hclient.request(req);
-        let www_auth = freq
-            .from_err()
-            .and_then(|r| {
+
+        let fres = self
+            .build_reqwest(reqwest::async::Client::new().get(url.clone()))
+            .send()
+            .map_err(|e| Error::from(format!("{}", e)))
+            .and_then(move |r| {
+                trace!("GET '{}' status: {:?}", r.url(), r.status());
                 let a = r
                     .headers()
-                    .get(hyper::header::WWW_AUTHENTICATE)
+                    .get(reqwest::header::WWW_AUTHENTICATE)
                     .ok_or_else(|| Error::from("get_token: missing Auth header"))?;
                 let chal = String::from_utf8(a.as_bytes().to_vec())?;
                 Ok(chal)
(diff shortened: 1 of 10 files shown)


src/v2/auth.rs
```diff
@@ -89,6 +85,16 @@ impl Client {
...
+        // TODO
+        trace!("login called");
+
+        self.login_hyper(scopes)
+    }
+
+    /// Perform registry authentication and return an authenticated token.
+    ///
+    /// On success, the returned token will be valid for all requested scopes.
+    pub fn login_hyper(&self, scopes: &[&str]) -> FutureTokenAuth {
```
lucab: This can stay private or `pub(crate)`.

src/v2/auth.rs
```diff
@@ -143,47 +149,65 @@ impl Client {
         Box::new(auth)
     }
 
+    /// Ensure the client is authenticated with the registry.
+    pub fn ensure_auth(&self, token: Option<&str>) -> FutureBool {
```
lucab: I think this is unused at the moment and can be dropped. If you are planning to use it in library consumer, I think it makes more sense to take ownership of `self` and return a `Future<Item = Self>`.

src/v2/tags.rs
```diff
@@ -20,53 +20,66 @@ struct TagsChunk {
...
+        let client = self.clone();
         let base_url = format!("{}/v2/{}/tags/list", self.base_url, name);
 
-        let fres = futures::stream::unfold(Some(String::new()), move |link| {
+        let fres = futures::stream::unfold(Some(String::new()), move |last| {
+            let client = client.clone();
+
             // Stream ends when response has no `Link` header.
-            let last = match link {
+            let last = match last {
```
lucab: Can we rename one of those two to avoid shadowing the name of the argument?

src/v2/tags.rs
```diff
@@ -20,53 +20,66 @@ struct TagsChunk {
...
+
+                    client
+                        .build_reqwest(reqwest::async::Client::new().get(url.clone()))
+                        .header(header::ACCEPT, "application/json")
+                        .send()
+                        .map_err(|e| Error::from(format!("{}", e)))
                 })
+                .flatten()
                 .and_then(|resp| {
+                    // ensure the status is OK
```
lucab: I think `reqwest` has a `error_for_status` which should implement this same logic.

src/v2/auth.rs
```diff
@@ -89,6 +85,16 @@ impl Client {
     ///
     /// On success, the returned token will be valid for all requested scopes.
     pub fn login(&self, scopes: &[&str]) -> FutureTokenAuth {
+        // TODO
```
lucab: I think you forgot the rest of the TODO message here.

src/v2/auth.rs
steveeJ: I'll remove the indirection for now as I haven't implemented the login in reqwest after all.

steveeJ:
@lucab comments addressed and tests fixed if the status goes green ;-)

steveeJ:
@lucab 
I can't reproduce the test freezes locally. Any idea what's different on Travis?

lucab:
@steveeJ not really, but I'm fine with disabling them for now to get this PR through.

PR #92, opened 2019-02-04
examples: improve authentication and argument handling

steveeJ:
* Only login when actually necessary, as not all registries seem to support authentication in the way we do it here.
* Improve the argument handling for some examples

### DIFF
--- a/examples/common/mod.rs
+++ b/examples/common/mod.rs
@@ -3,12 +3,12 @@ extern crate futures;
 
 use futures::prelude::*;
 
-pub fn authenticate_client<'a>(
-    client: &'a mut dkregistry::v2::Client,
-    login_scope: &'a str,
-) -> impl futures::future::Future<Item = &'a dkregistry::v2::Client, Error = dkregistry::errors::Error>
+pub fn authenticate_client(
+    mut client: dkregistry::v2::Client,
+    login_scope: String,
+) -> impl Future<Item = dkregistry::v2::Client, Error = dkregistry::errors::Error>
 {
-    futures::future::ok::<_, dkregistry::errors::Error>(client)
+    futures::future::ok::<_, dkregistry::errors::Error>(client.clone())
         .and_then(|dclient| {
             dclient.is_v2_supported().and_then(|v2_supported| {
                 if !v2_supported {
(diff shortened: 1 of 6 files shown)


examples/image.rs
```diff
@@ -13,14 +13,22 @@ use std::{boxed, env, error, fs, io};
...
+        None => "quay.io".into(),
+    };
 
-    println!("[{}] downloading image {}", registry, dkr_ref);
+    let image = match std::env::args().nth(2) {
+        Some(x) => x,
+        None => "coreos/etcd".into(),
+    };
+
+    let version = match std::env::args().nth(3) {
```
lucab: I'm not sure I'm getting the reason behind splitting registry / image / version here. Is reference parsing broken on some input?

steveeJ: It seems so: https://github.com/camallo/dkregistry-rs/pull/93

examples/common/mod.rs
```diff
@@ -6,9 +6,9 @@ use futures::prelude::*;
 pub fn authenticate_client<'a>(
     client: &'a mut dkregistry::v2::Client,
     login_scope: &'a str,
-) -> impl futures::future::Future<Item = &'a dkregistry::v2::Client, Error = dkregistry::errors::Error>
+) -> impl futures::future::Future<Item = dkregistry::v2::Client, Error = dkregistry::errors::Error> + 'a
```
lucab: I am not sure if I'm missing something, but it feels like this function could just take ownership of the input client and get rid of the lifetime. It is anyway cloning internally and returning a fully owned client on success.

examples/image.rs
lucab: Ack. I started massaging the parsing logic there to take care of such case, but let's keep this split here in the meanwhile.

examples/image.rs
```diff
@@ -54,20 +61,25 @@ fn main() {
...
 
 fn run(
-    dkr_ref: &reference::Reference,
+    registry: &str,
+    image: &str,
+    version: &str,
     user: Option<String>,
     passwd: Option<String>,
 ) -> Result<(), boxed::Box<error::Error>> {
+    env_logger::Builder::new()
```
lucab: I'm surprised this works (here and in other examples as well) as I don't see imports for either `env_logger` nor `log`. Do you know what's the magic trick here?

examples/common/mod.rs
lucab: Also, non-namespaced `Future` should be already in scope here as there is a prelude-use at the top-level.

examples/image.rs
steveeJ: I can't tell you. I went ahead and removed all `extern crate log/env_logger` from the examples and it still works.

lucab: It looks like this is an ergonomic improvement from 1.30 module rework: https://blog.rust-lang.org/2018/10/25/Rust-1.30.0.html#module-system-improvements. It works here because we are using the full (namespaced) path.

PR #93, opened 2019-02-04
reference: add test and extract constants

steveeJ:
Reference parsing is broken for inputs with ports and repos without a namespace.
This acts as an issue demonstration until we fix it.

@lucab please feel free top commit on top of this

### DIFF
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -32,6 +32,10 @@
 use std::str::FromStr;
 use std::{fmt, str};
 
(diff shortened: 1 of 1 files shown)


lucab:
I've split apart the constant and test chunks in here, and improved the parsing logic. I didn't want to force push here, so the result is at https://github.com/camallo/dkregistry-rs/pull/94.

steveeJ:
Superseded by #94

PR #94, opened 2019-02-05
reference: rework custom hostname handling

lucab:
This reworks the reference parser in order to better handle references with custom hostnames.

Supersedes: https://github.com/camallo/dkregistry-rs/pull/93


lucab:
@steveeJ thanks for splitting constants apart and for the testcase. I've kept those and reworked the parsing logic. If you think there is any corner-case label missing here, I'll be happy to add it.

lucab:
Rebased without any further functional changes. Tests are green. Keeping the LGTM and merging.

PR #96, opened 2019-02-08
cargo: expose reqwest TLS feature flags

lucab:
This exposes two underlying TLS feature flags from reqwest, allowing
consumers to pick openssl (default) or rustls implementation.

Closes https://github.com/camallo/dkregistry-rs/pull/86 (superseded).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,13 +30,16 @@ strum_macros = "0.13"
 tar = "0.4"
 tokio = "0.1"
 dirs = "1.0"
-reqwest = "0.9"
+reqwest = { version = "^0.9.6", default-features = false }
 
 [dev-dependencies]
 env_logger = "0.6"
 spectral = "0.6"
 
 [features]
+default = ["reqwest-default-tls"]
+reqwest-default-tls = ["reqwest/default-tls"]
+reqwest-rustls = ["reqwest/rustls-tls"]
 test-net = []
 test-mock = ["mockito"]
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,17 @@ conformant to the [Docker Registry HTTP API V2][registry-v2] specification.
 
 [registry-v2]: https://docs.docker.com/registry/spec/api/
 
+## Configurable features
+
+The following is a list of [Cargo features][cargo-features] that consumers can enable or disable:
+
+ * **reqwest-default-tls** *(enabled by default)*: provides TLS support via [system-specific library][native-tls] (OpenSSL on Linux)
+ * **reqwest-rustls**: provides TLS support via the [rustls][rustls] library
+
+[rustls]: https://docs.rs/rustls
+[native-tls]: https://docs.rs/native-tls
+[cargo-features]: https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-features-section
+
 ## Testing
 
 ### Integration tests
@@ -29,7 +40,7 @@ cargo test --features test-mock
 
 This library includes additional interoperability tests against some of the most common registries.
 
-Those tests are not run by default as they required network access and registry credentials.
+Those tests are not run by default as they require network access and registry credentials.
 
 They are gated behind a dedicated "test-net" feature and can be run as:
 ```


lucab:
I verified that re-exposing the feature flag works with:
```
$ cargo build --no-default-features --features reqwest-rustls  --example tags
$ ldd target/debug/examples/tags
        linux-vdso.so.1 (0x00007ffce4b8a000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5618d53000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f5618d49000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5618d28000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5618d0e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5618b4d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f56198af000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f56189ca000)
```

/cc @steveeJ

steveeJ:
Please explain the consequences for consumers after this change.

lucab:
@steveeJ do you want me to add a note to readme / docstring like https://docs.rs/reqwest/0.9.9/reqwest/#optional-features? Or are you just wondering about what's the default behavior?

steveeJ:
> do you want me to add a note to readme / docstring like https://docs.rs/reqwest/0.9.9/reqwest/#optional-features? Or are you just wondering about what's the default behavior?

The latter, I'm clueless ;-)

lucab:
Oh, sorry, let me go in more details here then.

reqwest recently gained feature flags to let consumers switch the underlying TLS implementation: https://docs.rs/reqwest/0.9.6/reqwest/#optional-features
Without further configuration, the [default set](https://github.com/seanmonstar/reqwest/blob/v0.9.9/Cargo.toml#L49) is `default-tls` (i.e. openssl). rustls implementation can be used instead by disabling the default one and activating `rustls-tls`.

This PR just re-exposes those options to consumers of dkregistry. This adds `reqwest-default-tls` in our default set, which in turns enables `default-tls` in reqwest (note the linkage against openssl libcrypto and libssl):
```
$ cargo build --example tags
$ ldd target/debug/examples/tags
        linux-vdso.so.1 (0x00007fff8a9e4000)
        libssl.so.1.0.2 => /usr/lib/x86_64-linux-gnu/libssl.so.1.0.2 (0x00007fbb20b66000)
        libcrypto.so.1.0.2 => /usr/lib/x86_64-linux-gnu/libcrypto.so.1.0.2 (0x00007fbb20901000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fbb208fc000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fbb208f2000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fbb208d1000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fbb208b7000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fbb206f4000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fbb214c0000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fbb20571000)
```

However, consumers can switch to rustls-only via the `reqwest-rustls` feature, in a way similar to reqwest flow (this embeds rustls in the binary and removes any openssl linkage):
```
$ cargo build --no-default-features --features reqwest-rustls  --example tags
$ ldd target/debug/examples/tags
        linux-vdso.so.1 (0x00007ffce4b8a000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5618d53000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f5618d49000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5618d28000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5618d0e000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5618b4d000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f56198af000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f56189ca000)
```

If a consumer doesn't care about any of this, not specifying any feature flag will result in reqwest default/historical behavior of just using openssl. https://github.com/openshift/cincinnati/pull/57 should fit in this case and not be concerned with rustls at all anymore.

I hope it is a bit clearer now, sorry for being terse initially. That said, I think we can still bikeshed on features name (should I stick to the same ones reqwest uses?) and see where to document them (is this comment enough or should I add a persistent note somewhere else?)

steveeJ:
Thanks for elaborating, very helpful and I understand the point now; we enable consumers of `dkregistry-rs` to configure the transitive dependency `reqwest`.

> That said, I think we can still bikeshed on features name (should I stick to the same ones reqwest uses?) and see where to document them (is this comment enough or should I add a persistent note somewhere else?)

I'm fine with the naming and the comment is enough for me to understand it, so it's likely enough for others.

lucab:
Ack, I pushed an additional commit on top to document these new feature flags.

PR #105, opened 2019-04-24
v2: add content digest module and verify blob integrity on download

steveeJ:
Motivation for this is to detect if a registry sends a blob which doesn't match its layer digest.

TODO
- [x] Unit tests for digest
- [x] Mock tests for get_blobs?

---

Depends on #104.

### DIFF
--- a/src/v2/blobs.rs
+++ b/src/v2/blobs.rs
@@ -36,24 +36,26 @@ impl Client {
 
     /// Retrieve blob.
     pub fn get_blob(&self, name: &str, digest: &str) -> FutureBlob {
-        let url = {
+        let fres_digest = futures::future::result(ContentDigest::try_new(digest.to_string()));
+
+        let fres_blob = {
             let ep = format!("{}/v2/{}/blobs/{}", self.base_url, name, digest);
-            match reqwest::Url::parse(&ep) {
-                Ok(url) => url,
-                Err(e) => {
-                    return Box::new(futures::future::err::<_, _>(Error::from(format!(
+            reqwest::Url::parse(&ep).map_err(|e|{
+                    ::errors::Error::from(format!(
                         "failed to parse url from string: {}",
                         e
-                    ))));
-                }
-            }
-        };
-
-        let fres = self.build_reqwest(reqwest::async::Client::new().get(url))
-            .send()
-            .map_err(|e| ::errors::Error::from(format!("{}", e)))
+                    ))
+            })
+            .map(|url|{
+                self.build_reqwest(reqwest::async::Client::new()
+                    .get(url))
+                    .send()
+                    .map_err(|e| ::errors::Error::from(format!("{}", e)))
+            })
+            .into_future()
+            .flatten()
             .and_then(|res| {
-                trace!("Blob GET status: {:?}", res.status());
+                trace!("GET {} status: {}", res.url(), res.status());
                 let status = res.status();
 
                 if status.is_success()
@@ -99,7 +101,14 @@ impl Client {
                         status
                     )))
                 }
-            });
+            })
+        };
+
+        let fres = fres_digest.join(fres_blob).and_then(|(digest, body)| {
+            digest.try_verify(&body)?;
+            Ok(body)
+        });
+
         Box::new(fres)
     }
 }
--- a/src/v2/content_digest.rs
+++ b/src/v2/content_digest.rs
@@ -1,0 +1,136 @@
+/// Implements types and methods for content verification
+use sha2::{self, Digest};
+use v2::*;
+
+/// ContentDigest stores a digest and its DigestAlgorithm
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct ContentDigest {
+    digest: String,
+    algorithm: DigestAlgorithm,
+}
+
+/// DigestAlgorithm declares the supported algorithms
+#[derive(Display, Clone, Debug, PartialEq, EnumString)]
+enum DigestAlgorithm {
+    #[strum(to_string = "sha256")]
+    Sha256,
+}
+
+impl ContentDigest {
+    /// try_new attempts to parse the digest string and create a ContentDigest instance from it
+    ///
+    /// Success depends on
+    /// - the string having a "algorithm:" prefix
+    /// - the algorithm being supported by DigestAlgorithm
+    pub fn try_new(digest: String) -> Result<Self> {
+        let digest_split = digest.split(':').collect::<Vec<&str>>();
+
+        if digest_split.len() != 2 {
+            return Err(format!("digest '{}' does not have an algorithm prefix", digest).into());
+        }
+
+        let algorithm =
+            std::str::FromStr::from_str(digest_split[0]).map_err(|e| format!("{}", e))?;
+        Ok(ContentDigest {
+            digest: digest_split[1].to_string(),
+            algorithm,
+        })
+    }
+
+    /// try_verify hashes the input slice and compares it with the digest stored in this instance
+    ///
+    /// Success depends on the result of the comparison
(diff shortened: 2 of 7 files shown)


src/v2/blobs.rs
```diff
@@ -1,11 +1,74 @@
...
 use reqwest;
 use reqwest::StatusCode;
+use sha2::{self, Digest};
 use v2::*;
 
 /// Convenience alias for future binary blob.
 pub type FutureBlob = Box<futures::Future<Item = Vec<u8>, Error = Error>>;
 
+#[derive(Clone, Debug, PartialEq)]
+pub struct LayerDigest {
```
lucab: Name bikeshedding: I think this isn't specific to layers, according to spec other objects (e.g. manifests) may return a `Docker-Content-Digest`. Spec calls this "content digest", I'd suggest to stick to that.

src/v2/blobs.rs
```diff
@@ -99,6 +164,10 @@ impl Client {
                         status
                     )))
                 }
+            }).
+            and_then(|body| {
+                layer_digest?.try_verify(&body)?;
```
lucab: Perhaps we should move `layer_digest?` earlier (or as a different input type for the function) and avoid hitting the network if we already know the digest is invalid.

src/v2/blobs.rs
steveeJ: I was going to have a discussion around this. I would agree that this can be used elsewhere and I want to find a reusable place for this in the internal API.

steveeJ: Moved and renamed into a separate module

src/v2/blobs.rs
steveeJ: I moved it into a separate future and serialized the two.

steveeJ:
@lucab All your comments are addressed, PTAL. Also, do you think mock tests would make sense for `get_blob`? We only have them for `has_blob` at the moment.

lucab:
@steveeJ I'm fine with either a mock-test or making sure travis runs a get_blob+verify against a live registry (I didn't check, I hope it does already). If you care about ensuring we catch the negative case (received blob with non-matching digest), then I think a mock-test is the only way.

Except for tests, are there any other WIP parts in here?

steveeJ:
> If you care about ensuring we catch the negative case (received blob with non-matching digest), then I think a mock-test is the only way.

I'll give mocking this a try, shouldn't be too complex.

> Except for tests, are there any other WIP parts in here?

Except for the test and forming a proper git history I consider it complete.

steveeJ:
@lucab I added the most tests for `get_blob`. I consider this ready for final review. That said, I'm open to discussing the stronger typing around the digest again.

src/v2/blobs.rs
```diff
@@ -100,6 +102,12 @@ impl Client {
                     )))
                 }
             });
+
+        let fres = fres_digest.join(fres_reqwest).and_then(|(digest, body)| {
```
lucab: `s/join/and_then/` maybe? As in, don't get to `reqwest` if the input-digest is known to be malformed?

src/v2/blobs.rs
```diff
@@ -49,11 +49,13 @@ impl Client {
             }
         };
 
-        let fres = self.build_reqwest(reqwest::async::Client::new().get(url))
+        let fres_digest = futures::future::result(ContentDigest::try_new(digest.to_string()));
```
lucab: I fear this should be moved even earlier. This line does input validation for `digest`, but we already use `digest` for format+parse. It should logically precede that.

If the url-parsing becomes too unwieldy  because of that, you can consider refactoring into a chain like https://github.com/camallo/dkregistry-rs/blob/0d4317d0e9816568d974d94570a7b87d3230f20e/src/v2/mod.rs#L97-L105

src/v2/content_digest.rs
```diff
@@ -0,0 +1,136 @@
...
+/// ContentDigest stores a digest and its DigestAlgorithm
+#[derive(Clone, Debug, PartialEq)]
+pub struct ContentDigest {
+    digest: String,
+    algorithm: DigestAlgorithm,
+}
+
+/// DigestAlgorithm declares the supported algorithms
+#[derive(Display, Clone, Debug, PartialEq, EnumString)]
+pub enum DigestAlgorithm {
```
lucab: This is not exposed in the library API, I think you can remove the `pub` (or scope it down to `pub(crate)`).

src/v2/content_digest.rs
```diff
@@ -0,0 +1,136 @@
...
+        let digest_split = digest.split(':').collect::<Vec<&str>>();
+
+        if digest_split.len() != 2 {
+            return Err(format!("digest '{}' does not have an algorithm prefix", digest).into());
+        }
+
+        let algorithm =
+            std::str::FromStr::from_str(digest_split[0]).map_err(|e| format!("{}", e))?;
+        Ok(ContentDigest {
+            digest: digest_split[1..].join(""),
```
lucab: I'm not sure I follow the `[1..].join("")` part. We checked above that we have exactly two fields, so I would expect a `[1]` here, no?

src/v2/content_digest.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// DigestAlgorithm declares the supported algorithms
+#[derive(Display, Clone, Debug, PartialEq, EnumString)]
+pub enum DigestAlgorithm {
+    #[strum(to_string = "sha256")]
+    Sha256,
+}
+
+impl ContentDigest {
+    /// try_new attempts to parse the digest string and create a ContentDigest instance from it.error_chain
```
lucab: Spurious trailing `error_chain`.

src/v2/content_digest.rs
```diff
@@ -0,0 +1,136 @@
...
+    ///
+    /// Success depends on
+    /// - the string having a "algorithm:" prefix
+    /// - the algorithm being supported by DigestAlgorithm
```
lucab: Do we want to add a "the digest is a valid value for DigestAlgorithm"? I think right now this accepts things like "sha256:thanks4allthefish".

src/v2/blobs.rs
steveeJ: While looking at the documented of `join` is not fully conclusive, [it is documented to drop the other future if one fails](https://docs.rs/futures/0.1.26/futures/future/trait.Future.html#method.join). Exploring your suggestion, the `join` can't be simply substituted by `and_then` which expects an `FnOnce -> IntoFuture` (simplified) as its argument as opposed to an `IntoFuture`, so it would complicate things more by nesting a level.

Looking at the macro which generates the `join` method, reveals that it [polls the operands in the given order and fails early](https://github.com/rust-lang-nursery/futures-rs/blob/043b16f96732c07df560e6198728ec470d7e96e6/src/future/join.rs#L66-L82), so I think we're good here.

src/v2/content_digest.rs
steveeJ: I thought about that too. We'd have one regular expression per algorithm to verify the digest further. I don't thing it's our responsibility though. If you strongly feel we should do it I'll incorporate it

src/v2/blobs.rs
lucab: Ack, let's stick to `join` then.

src/v2/content_digest.rs
lucab: Not a strong feeling, I'm fine with keeping this as is for this PR.

lucab:
#104 landed in the meanwhile.

steveeJ:
> #104 landed in the meanwhile.

I'm pretty sure the commit was already in this history, but I rebased again and now the merge commit is in here too. I can see a lot of conversations which aren't marked as resolved. I think they're all settled though, please take a look if I missed anything.

PR #107, opened 2019-04-26
[WIP] v2: verify manifest digest

steveeJ:
This adds conditional manifest digest verification. 
If the manifest is download by reference and the registry provides a digest in the response they have to be the same. Additionally the calculated digest of the downloaded manifest has to match the requested one

TODO
- [ ] handle differences in v2s1 and v2s2 when calculating the digest of the downloaded manifest
- [ ] tests

--- 

Depends on #105.


lucab:
#105 is now merged.

PR #112, opened 2019-09-04
v2/auth: properly parse bearer strings with CSVs

steveeJ:
As encountered on a Satellite instance the bearer values may be in CSV
notation, which was previously broken in the auth code.

The CSV notation conforms to the [spec][0], which defines:

```
resourcescope := resourcetype  ":" resourcename  ":" action [ ',' action ]*
```

[0]: https://docs.docker.com/registry/spec/auth/scope/

---

/cc @vrutkovs FYI

### DIFF
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -50,17 +50,8 @@ impl Client {
                 Ok(chal)
             })
             .and_then(move |hdr| {
-                let mut auth_ep = "".to_owned();
-                let mut service = None;
-                for item in hdr.trim_start_matches("Bearer ").split(',') {
-                    let kv: Vec<&str> = item.split('=').collect();
-                    match (kv.get(0), kv.get(1)) {
-                        (Some(&"realm"), Some(v)) => auth_ep = v.trim_matches('"').to_owned(),
-                        (Some(&"service"), Some(v)) => service = Some(v.trim_matches('"')),
-                        (Some(&"scope"), _) => {}
-                        (key, _) => return Err(format!("unsupported key '{:?}'", key).into()),
-                    };
-                }
+                let (mut auth_ep, service) = parse_hdr_bearer(hdr.trim_start_matches("Bearer "))?;
+
                 trace!("Token provider: {}", auth_ep);
                 if let Some(sv) = service {
                     auth_ep += &format!("?service={}", sv);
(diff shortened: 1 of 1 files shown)


src/v2/auth.rs
```diff
@@ -185,3 +176,54 @@ impl Client {
         Box::new(fres)
     }
 }
+
+// This parses a Www-Authenticate header of value Bearer, which could look something like this:
+// `Www-Authenticate: Bearer realm="https://sat-r220-02.lab.eng.rdu2.redhat.com/v2/token",service="sat-r220-02.lab.eng.rdu2.redhat.com",scope="repository:registry:pull,push"`
```
lucab: I'd put the spec URL here for reference, move this example to the test, and just mention "which looks like the test sample".
Also, can you please turn this into a full docstring comment with `///` and title+body split?

src/v2/auth.rs
```diff
@@ -185,3 +176,54 @@ impl Client {
...
+
+    let re = regex::Regex::new(r#"(([a-z]+)="([^"]*)")"#)?;
+    for capture in re.captures_iter(input) {
+        match (
+            // The indices for this capture are as follows:
+            // 0: full match
+            // 1: outer group match
+            // 2: first nested group match
+            // 3: second nested group match
+            // Hence, we are interested in the sub-group matches i.e. in 2 and 3.
```
lucab: Thanks for the detailed comment. Just personal taste, but the match with nested comment looks funny.
What about naming those two variables (key+value?) and moving them plus comments outside of the match body?

src/v2/auth.rs
steveeJ: I acknowledge that it's pretty verbose. Since the test already has this string I can just remove it here. Alternatively we could do a docstring test. How about a doctest for this one? It seems small enough

lucab: Interesting, I guess we can do that.
But I thought you wanted to show the whole `Www-Authenticate: Bearer realm=<foo>` thing and not just `foo`, right?

steveeJ: Not necessarily. The difficulty is caused by the stuff after `Bearer ` so I'm fine leaving the the trimming out of the example.

steveeJ:
Thanks for the quick review @lucab. Comments are addressed, PTAL.

Issue #83, opened 2019-01-14
v2/client: trim content-type parameters


lucab:
This is a followup to https://github.com/camallo/dkregistry-rs/pull/82#issuecomment-453737148.

Current docker-registry specification and implementation have some quirks related to content-type headers, returning charset parameters even for media-types that don't allow any.

As a minor compatibility fix, we should revert https://github.com/camallo/dkregistry-rs/pull/82 and instead introduce a `mime_trim_params` that expunges all parameters from a media type.

schrieveslaach:
@lucab, I noticed that there are also errors with Nexus 3 (see aixigo/prevant#44). Is this issue related to it?

lucab:
@schrieveslaach no, your troubles seem unrelated to this specific issue.
I think you are just hitting some content-negotiation bug, and it wouldn't be the first registry we encounter with some buggy logic.
It would be helpful to have a minimal Rust example to look at and a trace at the HTTP level.

schrieveslaach:
@lucab, sorry for interfering. [It is an issue of Nexus 3](https://issues.sonatype.org/browse/NEXUS-24124).

PR #113, opened 2019-09-05
mediatyeps: detect utf-8 encoded JSON manifest

steveeJ:
This is returned by the Satellite registry.


lucab:
This is a bug in Satellite, as `application/json` is a binary format and thus doesn't allow a textual charset.
This was already discussed, covered and fixed in https://github.com/opencontainers/distribution-spec/pull/41.

While I don't have problems landing this PR as a temporary fix, I would really like to first have a ticket to track on Satellite side to fix it (and some timebox to remove this hack).

steveeJ:
I filed a BZ ticket for this: https://bugzilla.redhat.com/show_bug.cgi?id=1749317

steveeJ:
I think I misunderstood your comment. Do I understand correctly now that the content-type is actually contradicting in itself?

lucab:
> Do I understand correctly now that the content-type is actually contradicting in itself?

Yes. `application/json` takes no additional parameters, as per RFC.

That said, I'm fine with either landing this band-aid or do https://github.com/camallo/dkregistry-rs/issues/83 and relax the parser, as long Satellite gets fixed at some point.

PR #114, opened 2019-09-05
refactor login logging and example

steveeJ:


src/v2/auth.rs
```diff
@@ -130,8 +130,24 @@ impl Client {
...
+            .and_then(|token_auth: TokenAuth| {
+                let mut t = token_auth.token().to_string();
+
+                if t == "unauthenticated" {
+                    bail!("received token with value '{}'", t)
+                } else if t.is_empty() {
+                    bail!("received an empty token")
+                };
+
+                // mask the token before logging it
```
lucab: If we like minimalism, I'm fine with just logging the token length here instead.

src/v2/auth.rs
```diff
@@ -151,22 +167,15 @@ impl Client {
...
+        let req = self.build_reqwest(reqwest::async::Client::new().get(url.clone()));
+        let req = if let Some(t) = token {
+            req.bearer_auth(t)
         } else {
             debug!("is_auth called without token");
+            req
         };
 
-        trace!("Sending reqwest '{:?}'", req);
+        trace!("Sending reqwest to '{}'", url);
```
lucab: s/reqwest/request/ I guess?

steveeJ: Aren't they synonymous now in the context of Rust programs? :wink:

src/v2/auth.rs
steveeJ: If you're not opposed to the masked string I'd opt for leaving it in. Please make the call

lucab: Code is already there, let's keep it.

PR #117, opened 2019-09-23
v2: introduce a common Manifest type and refactor related functions

steveeJ:
This type serves as an abstraction for the various manifest types and
makes it easier for consumers to retrieve information from it.


src/v2/manifest/manifest_schema2.rs
```diff
@@ -9,6 +14,8 @@ pub struct ManifestSchema2 {
     media_type: String,
     config: Config,
     layers: Vec<S2Layer>,
+    #[serde(skip)]
+    config_blob: Option<ConfigBlob>,
```
lucab: Why is this field skipped?

steveeJ: It's not part of any manifest spec. The intention is to fill this option in with the `fetch_config_blob` method.

lucab: Mmhhh, that smells quite a bit as we are now mixing together data structures with external specification and internal data-types.
Would it be possible to keep the `ManifestSchema2` and `ConfigBlob` separate and have a super-type which embeds both?

src/v2/manifest/manifest_schema2.rs
```diff
@@ -19,6 +26,12 @@ pub struct Config {
     digest: String,
 }
 
+/// Partial representation of a container image (application/vnd.docker.container.image.v1+json)
```
lucab: Is the full version of this described somewhere?

lucab:
@steveeJ if quay.io already supports this on some public repo that we can use, then I think we can simply use that for integration tests.

src/v2/manifest/manifest_schema2.rs
steveeJ: I didn't find it fast enough when I was implementing it, so I just picked the field which I was interested at the time.

src/v2/manifest/manifest_schema2.rs
steveeJ: Probably so ;-) Would you prefer that?

lucab: Yes, if the resulting API does not turn into a spaghetti dish of types. I think it will help in the longer to keep the external data-structure as easy to cross-check as possible.

steveeJ: I introduced a container type which embeds the manifest and the blob. It looks reasonable to me. PTAL

src/v2/manifest/manifest_schema2.rs
steveeJ: I want to add that this was meant as an RFC, and I'll gladly implement the whole structure instead of just the field, and also reference the spec for it ;-)

src/v2/manifest/manifest_schema2.rs
```diff
@@ -11,12 +16,25 @@ pub struct ManifestSchema2 {
     layers: Vec<S2Layer>,
 }
 
-#[derive(Debug, Default, Deserialize, Serialize)]
+/// Super-type for combining a ManifestSchema2 with a ConfigBlob
```
lucab: Nit: missing trailing dot.

src/v2/manifest/manifest_schema2.rs
lucab: I think this is described here https://github.com/moby/moby/blob/master/image/spec/v1.md#image-json-description.

src/v2/manifest/manifest_schema2.rs
```diff
@@ -61,16 +79,82 @@ pub struct Platform {
     pub features: Option<Vec<String>>,
 }
 
+impl ManifestSchema2Spec {
+    /// Get digest of the configuration object referenced by this manifest.
```
lucab: This does not return the digest, though. In which direction do you want to fix it?

src/v2/manifest/mod.rs
```diff
@@ -29,7 +32,7 @@ impl Client {
     ///
     /// The name and reference parameters identify the image.
     /// The reference may be either a tag or digest.
-    pub fn get_manifest_and_ref(&self, name: &str, reference: &str) -> FutureManifestAndRef {
+    pub fn get_manifest_and_ref(&self, name: String, reference: &str) -> FutureManifestAndRef {
```
lucab: Can we keep the input type consistent between this and `get_manifest` (in whichever direction you prefer)?

src/v2/manifest/mod.rs
```diff
@@ -77,22 +89,59 @@ impl Client {
...
+                    .and_then(move |m| {
+                        m.fetch_config_blob(client_spare0, name.to_string())
+                            .map(Manifest::S2)
+                    })),
+                    mediatypes::MediaTypes::ManifestList => A(futures::future::result(
+                        serde_json::from_slice::<ManifestList>(&body)
+                            .map_err(|e| Error::from(format!("{}", e)))
+                            .map(Manifest::ML),
+                    )),
+                    unsupported => A(futures::future::result(Err(Error::from(format!(
```
lucab: You can directly use `future::result` here.

src/v2/manifest/mod.rs
```diff
@@ -1,10 +1,13 @@
 //! Manifest types.
 use mediatypes;
+use std::collections::HashMap;
 use v2::*;
 
+use futures::future::Either::{A, B};
```
lucab: For readability, I'd keep this use at the `Either` level.

src/v2/manifest/mod.rs
```diff
@@ -234,3 +260,100 @@ fn to_mimes(v: &[&str]) -> Result<Vec<mime::Mime>> {
...
+                self
+            )
+            .into()),
+        }
+    }
+
+    /// The architecture of the image the manifest points to, if available.
+    ///
+    /// For some Manifest types this could result in an external request, hence
+    /// the result is wrapped in an asynchronous type.
```
lucab: This docstring does not seem to match the signature.

src/v2/manifest/mod.rs
```diff
@@ -234,3 +260,100 @@ fn to_mimes(v: &[&str]) -> Result<Vec<mime::Mime>> {
...
+            _ => Err(format!(
+                "Manifest {:?} doesn't support the 'architecture' method",
+                self
+            )
+            .into()),
+        }
+    }
+
+    /// Get a collection of all image labels for the given layer, if available.
+    pub fn labels(&self, layer: usize) -> Result<Option<(HashMap<String, String>)>> {
```
lucab: I have the feeling that this method does not belong here. Where are the labels for Schema2 and ManifestList types?

src/v2/manifest/manifest_schema2.rs
steveeJ: Would you  prefer to have the full struct implemented for completeness?

src/v2/manifest/mod.rs
steveeJ: Ah, I think you meant `future::err`, right? Thanks in any case for the tip

src/v2/manifest/mod.rs
steveeJ: I had the same feeling. I'm fine dropping it, since I'm also not aware of any consumers.

src/v2/manifest/mod.rs
lucab: Oh, even better :+1:

src/v2/manifest/mod.rs
lucab: Let's drop it then (assuming Cincinnati is not using it).

src/v2/manifest/manifest_schema2.rs
lucab: I think that's overkill in this scope. Just link to the spec and leave a note that there are remaining fields that could be added in the future.

steveeJ: Thanks for digging for the spec. It's referenced int the docstring now.

src/v2/manifest/manifest_schema2.rs
```diff
@@ -61,16 +83,82 @@ pub struct Platform {
...
+            .layers
+            .iter()
+            .map(|l| l.digest.clone())
+            .collect()
     }
 
-    /// Get digest of the configuration object referenced by this manifest.
-    pub fn config(&self) -> String {
-        self.config.digest.clone()
+    /// Get the architecture from the config_blob
```
wking: `config_blob` is just internal, so this could be `from the config.`

src/v2/manifest/mod.rs
```diff
@@ -48,22 +52,31 @@ impl Client {
...
-            Err(e) => {
-                let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                error!("{}", msg);
-                return Box::new(futures::future::err::<_, _>(Error::from(msg)));
-            }
-        };
+        let accept_headers = header::HeaderMap::from_iter(
+            [
+                mediatypes::MediaTypes::ManifestV2S2,
+                mediatypes::MediaTypes::ManifestV2S1Signed,
```
wking: Probably want to set [`q`][1] for these so we're making our preferences clear.

[1]: https://tools.ietf.org/html/rfc7231#section-5.3.2

src/v2/manifest/mod.rs
```diff
@@ -238,3 +270,87 @@ fn to_mimes(v: &[&str]) -> Result<Vec<mime::Mime>> {
...
+
+impl Manifest {
+    /// List digests of all layers referenced by this manifest, if available.
+    ///
+    /// The returned layers list is ordered starting with the base image first.
+    pub fn layers_digests(&self) -> Result<Vec<String>> {
+        match self {
+            Manifest::S1Signed(m) => Ok(m.get_layers()),
+            Manifest::S2(m) => Ok(m.get_layers()),
+            // Manifest::ML(_) => TODO(steveeJ),
```
wking: To get layers out of a manifest-list, you need to chose an architecture so you know which referenced digest to descend into.  If the API is `layers_digests(&self, architecture: Option<String>) -> ...`, then you can do that.  If the option is passed, you want a check for the `S1Signed` and `S2` that looks into the config to see if the incoming architecture matches the value set there.  In the unset-option case, there is only one choice (`S1Signed` and `S2`, but also `ML` if there happens to only be a single architecture).

src/v2/manifest/mod.rs
```diff
@@ -238,3 +270,87 @@ fn to_mimes(v: &[&str]) -> Result<Vec<mime::Mime>> {
...
+            _ => Err(format!(
+                "Manifest {:?} doesn't support the 'layer_digests' method",
+                self
+            )
+            .into()),
+        }
+    }
+
+    /// The architecture of the image the manifest points to, if available.
+    pub fn architecture(&self) -> Result<String> {
```
wking: This should probably return a set, which could be empty (e.g. [`manifests` can be an empty-array][1] for the very-similar OCI indexes).  I don't care if we throw an error on that case or return an empty set, because OpenShift consumers will only care about images with architectures (there's no point to have OpenShift consumers look at an image that has no architecture).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.1/image-index.md#L26

src/v2/manifest/mod.rs
```diff
@@ -77,22 +90,68 @@ impl Client {
...
+                    header_content_type,
+                    media_type
+                );
+
+                Ok((body, content_digest, media_type))
+            })
+            .and_then(move |(body, content_digest, media_type)| {
+                match media_type {
+                    mediatypes::MediaTypes::ManifestV2S1Signed => {
+                        Either::A(futures::future::result(
```
vrutkovs: Its not clear why its different from `ManifestV2S2` code.

Shouldn't it be
```
mediatypes::MediaTypes::ManifestV2S1Signed => Either::A(
	futures::future::result(serde_json::from_slice::<ManifestSchema1Signed>(
		&body,
	))
	.map_err(|e| Error::from(format!("{}", e)))
...
```
?

steveeJ: Are you wondering about the curly braces?

vrutkovs: Yeah, pretty much. This is probably very minor issue, just a bit confusing if these cases are similar

steveeJ: The braces are added back by `cargofmt` if I remove them; that settles the styling issue.
The `ManifestV2S2` case has the additional step of downloading the config blob, which is a non-immediate future. compared to the `ManifestV2S1` case which only has the immediate value inside the future. That's why they are different as in `Either::A` and `Either::B`.

steveeJ:
@wking I addressed all of your comments. PTAL

src/v2/manifest/mod.rs
```diff
@@ -238,3 +274,101 @@ fn to_mimes(v: &[&str]) -> Result<Vec<mime::Mime>> {
...
+    S1Signed(manifest_schema1::ManifestSchema1Signed),
+    S2(manifest_schema2::ManifestSchema2),
+    ML(manifest_schema2::ManifestList),
+}
+
+impl Manifest {
+    /// List digests of all layers referenced by this manifest, if available.
+    ///
+    /// The returned layers list is ordered starting with the base image first.
+    pub fn layers_digests(&self, archtecture: Option<&str>) -> Result<Vec<String>> {
```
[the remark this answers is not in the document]
wking: > thanks, fixed!

Forgot to push?

```console
$ git fetch origin
$ git describe origin/pr/117
0.2.1-133-g7ac9ded
$ git grep archtecture origin/pr/117
origin/pr/117:src/v2/manifest/mod.rs:    pub fn layers_digests(&self, archtecture: Option<&str>) -> Result<Vec<String>> {
origin/pr/117:src/v2/manifest/mod.rs:        match (self, self.architectures(), archtecture) {
```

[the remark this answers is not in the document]
steveeJ: No :thinking: 

```console
$ git describe pr/verify-manifest-digests
0.2.1-103-ge10f549
steveej@steveej-t480s-work ✓ ~/src/job-redhat/dkregistry-rs  pr/refactor-mediatypes ⚡
$ git push -f
Everything up-to-date
```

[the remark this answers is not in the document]
wking: This PR uses pr/refactor-mediatypes, not pr/verify-manifest-digests.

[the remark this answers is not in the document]
steveeJ: That explains the `git describe` mismatch I'm seeing between our outputs :smile: Apologies!

But, it doesn't explain your suspicion that something's unpushed

```console
$ git describe pr/refactor-mediatypes
0.2.1-133-g8cb9d95
steveej@steveej-t480s-work ✓ ~/src/job-redhat/dkregistry-rs  pr/refactor-mediatypes ⚡
$ git describe origin/pr/refactor-mediatypes
0.2.1-133-g8cb9d95
```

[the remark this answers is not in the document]
steveeJ: Issue solved. I had replaced one typo with another, apologies again :wink:

PR #120, opened 2019-10-23
v2: add `Send` bound to most of the API

steveeJ:
This is required for consumers who want to use these methods in a
multi-threaded program.

Along the way some error conversion had to be changed.

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -162,7 +162,7 @@ impl Client {
             .build_reqwest(reqwest::async::Client::new().get(url.clone()))
             .headers(accept_headers)
             .send()
-            .map_err(|e| Error::from(format!("{}", e)))
+            .map_err(Error::from)
             .inspect(move |_| {
                 trace!("HEAD {:?}", url);
             })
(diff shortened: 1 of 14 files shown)


src/v2/manifest/mod.rs
```diff
@@ -162,15 +162,18 @@ impl Client {
             .build_reqwest(reqwest::async::Client::new().get(url.clone()))
             .headers(accept_headers)
             .send()
-            .map_err(|e| Error::from(format!("{}", e)))
+            .map_err(Error::from)
             .inspect(move |_| {
                 trace!("HEAD {:?}", url);
             })
-            .and_then(|r| {
+            .and_then(|r| -> std::result::Result<_, ::errors::Error> {
```
lucab: Isnt't this return type the same as `::errors::Result<_>`?
(I'm not sure why this needs additional hinting, but I trust that you fought the compiler before ending up with this)

steveeJ: This was part of debugging and leftover. Thanks for catching!

steveeJ:
@lucab I addressed your comment, and also added a commit on top to address all compiler warnings. PTAL and let me know if you'd prefer a separate PR for that or it is fine here.

PR #121, opened 2019-10-24
reference: add parse_url tests

steveeJ:
@lucab I think something's wrong with the `reference::parse_url` function, or do I have the wrong expectations?

### DIFF
--- a/src/reference.rs
+++ b/src/reference.rs
@@ -169,6 +169,27 @@ fn parse_url(input: &str) -> Result<Reference, Error> {
         .map(String::from)
         .collect();
 
+    // Figure out if the first component is a registry String, and assume the
+    // default registry if it's not.
+    let first = components
+        .pop_front()
+        .ok_or(Error::from("missing image name"))?;
+
+    let registry = if regex::Regex::new(r"(?x)
+        ^
+        # hostname
+        (([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)+([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9])
+
+        # optional port
+        ([:][0-9]{1,6})?
+        $
+    ")?.is_match(&first) {
+        first
+    } else {
+        components.push_front(first);
+        DEFAULT_REGISTRY.to_string()
+    };
+
     // Take image name and extract tag or digest-ref, if any.
     let last = components
         .pop_back()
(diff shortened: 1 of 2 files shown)


lucab:
@steveeJ I'm surprised too, as the cases in `tests/reference.rs` seems to be green. Perhaps we regressed this specific case in https://github.com/camallo/dkregistry-rs/pull/94?

steveeJ:
@lucab I added more test cases and an attempt to resolve the issue. PTAL

tests/reference.rs
```diff
@@ -7,23 +7,31 @@ use std::str::FromStr;
...
+    fn run_testcases(tcases: &[&str], expected: &str) {
+        for t in tcases {
+            let r = Reference::from_str(t);
+            asserting(t).that(&r).is_ok();
+            let repo = r.unwrap().repository();
+            asserting(t).that(&repo.as_str()).is_equal_to(expected);
+        }
     }
+    run_testcases(
+        &[
```
lucab: Can you please split these testcases further in two blocks, so that we check the first half gets the default registry while the second half gets custom values?

src/reference.rs
```diff
@@ -169,6 +169,27 @@ fn parse_url(input: &str) -> Result<Reference, Error> {
         .map(String::from)
         .collect();
 
+    // Figure out if the first component is a registry String, and assume the
+    // default registry if it's not.
+    let first = components
+        .pop_front()
+        .ok_or(Error::from("missing image name"))?;
+
+    let registry = if regex::Regex::new(r"(?x)
```
lucab: Mmmh, maybe a `format!("^{hostname}{opt_port}$")` is a bit more readable instead of inline comments? Also, what's the leading `(?x)` for?

src/reference.rs
```diff
@@ -184,23 +205,27 @@ fn parse_url(input: &str) -> Result<Reference, Error> {
 
     // Handle images in default library namespace, that is:
     // `ubuntu` -> `library/ubuntu`
-    if components.is_empty() {
+    if components.is_empty() && registry == DEFAULT_REGISTRY.to_string() {
```
lucab: Nit: maybe `&registry ==` instead?

src/reference.rs
steveeJ: Hehe, I had the format variant previously but didn't like it as much. If you feel strongly about it I'll revert back.

tests/reference.rs
steveeJ: I thought that's exactly what I did, isn't that so?

steveeJ: Oh, registry, I assumed repository.

src/reference.rs
steveeJ: >  what's the leading `(?x)` for?

http://www.rexegg.com/regex-modifiers.html#freespacing

lucab: Ack, I'm fine with keeping it as is.

PR #133, opened 2019-12-20
Switch to Rust edition 2018

steveeJ:
Based on #136.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ exclude = [
 ".gitignore",
 ".travis.yml",
 ]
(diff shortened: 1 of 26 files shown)


lucab:
Travis is not happy about that, but I think it's a matter of fixing a few imports.

If you haven't already, please `cargo fmt` this as well at the end.

steveeJ:
> Travis is not happy about that, but I think it's a matter of fixing a few imports.

I forgot to run the tests with the `test-mock` and `test-net` features. For some reason a network test times out locally, let's see what it does on travis.

PR #134, opened 2019-12-21
travis: cache cargo

steveeJ:
Based on #136

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ rust:
   - stable
   - beta
   - nightly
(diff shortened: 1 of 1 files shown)


lucab:
This looks fine to me, but I'm unsure what's the rationale. Is this only meant to speed up CI runs (and is it good at that)?

steveeJ:
I assume it'll save energy and also some time, but we'll see. Since it's a native Travis feature I assume it's well tested and we can safely give it a try.

PR #138, opened 2020-01-02
use async/await and update http, tokio

schrieveslaach:


schrieveslaach:
@lucab, @steveeJ, this PR adds support for async/await. Please, have a look.

Cargo.toml
```diff
@@ -25,27 +25,27 @@ tag-prefix = ""
...
 regex = "^1.1.0"
-serde = "1"
-serde_derive = "1"
+serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 strum = "0.16"
 strum_macros = "0.16"
 tar = "0.4"
-tokio = "0.1"
+tokio = "0.2"
```
schrieveslaach: Should I bump the minor version as well or do you update it?

Cargo.toml
```diff
@@ -25,27 +25,27 @@ tag-prefix = ""
...
 dirs = "2.0"
-reqwest = { version = "^0.9.6", default-features = false }
+reqwest = { version = "0.10", default-features = false, features = ["json"] }
 sha2 = "^0.8.0"
 
 [dev-dependencies]
 env_logger = "0.7"
 spectral = "0.6"
+tokio = { version = "0.2", features = ["macros"] }
```
lucab: To the best of my knowledge, cargo will do a union of the features-set between this and the other `tokio` dependency.

You can either promote the `macros` feature to the other dependency, or avoid using macros in examples (my preference is for the latter). At that point, you can simply drop this line.

src/errors.rs
```diff
@@ -4,7 +4,6 @@ use base64;
 use http;
 use regex;
 use reqwest;
-use serde_json;
```
lucab: Thanks for cleaning this up. I think you can drop the other imports too.

src/lib.rs
```diff
@@ -8,37 +8,36 @@
...
 //! };
 //! #
 //! # Ok(())
 //! # };
-//! # run().unwrap();
+//! # run().await.unwrap();
 //! # }
 //! ```
 
+#![feature(async_closure)]
```
lucab: This is an unstable feature, which doesn't work on stable toolchains. Please drop it.

Cargo.toml
lucab: I'm not sure I get your question, but feel free to keep the bump in this PR.

src/lib.rs
steveeJ: For reference: https://github.com/rust-lang/rust/issues/62290

src/v2/catalog.rs
```diff
@@ -1,19 +1,19 @@
...
+pub type StreamCatalog<'a> = BoxStream<'a, Result<String>>;
 
 #[derive(Debug, Default, Deserialize, Serialize)]
 struct Catalog {
     pub repositories: Vec<String>,
 }
 
 impl v2::Client {
-    pub fn get_catalog(&self, paginate: Option<u32>) -> StreamCatalog {
+    pub fn get_catalog<'a, 'b: 'a>(&'b self, paginate: Option<u32>) -> StreamCatalog<'a> {
```
steveeJ: When I started to take at look at porting dkregistry to async/.await, I was pointed towards https://github.com/tokio-rs/async-stream for stream handling. Do you by chance have experience with it, or are otherwise willing to compare it to what's currently in this PR?

src/v2/mod.rs
```diff
@@ -42,15 +38,14 @@ mod catalog;
 pub use self::catalog::StreamCatalog;
 
 mod auth;
-pub use self::auth::{FutureTokenAuth, TokenAuth};
+pub use self::auth::{TokenAuth};
```
steveeJ: superfluous brackets

tests/mock/api_version.rs
```diff
@@ -3,7 +3,7 @@ extern crate mockito;
 extern crate tokio;
 
 use self::mockito::mock;
-use self::tokio::runtime::current_thread::Runtime;
```
steveeJ: Out of curiosity, is there an equivalent in tokio 0.2 to restricting the runtime to the current thread?

tests/mock/catalog.rs
```diff
@@ -70,15 +74,15 @@ fn test_catalog_paginate() {
...
-    let (page1, next) = runtime.block_on(next.into_future()).ok().unwrap();
-    assert_eq!(page1, Some("r1/i1".to_owned()));
+    let (page1, next) = runtime.block_on(next.into_future());
+    assert_eq!(page1.unwrap().unwrap(), "r1/i1".to_owned());
 
-    let (page2, next) = runtime.block_on(next.into_future()).ok().unwrap();
+    let (page2, next) = runtime.block_on(next.into_future());
     // TODO(lucab): implement pagination
-    assert_eq!(page2, None);
+    assert!(page2.is_none());
```
steveeJ: This change makes the panic less helpful because it hides the value which was received.

tests/mock/catalog.rs
```diff
@@ -70,15 +74,15 @@ fn test_catalog_paginate() {
...
 
-    let (end, _) = runtime.block_on(next.into_future()).ok().unwrap();
-    assert_eq!(end, None);
+    let (end, _) = runtime.block_on(next.into_future());
+    assert!(end.is_none());
```
steveeJ: same as above

tests/mock/tags.rs
```diff
@@ -75,14 +76,14 @@ fn test_tags_paginate() {
...
 
-    let (second_tag, stream_rest) = runtime.block_on(stream_rest.into_future()).ok().unwrap();
-    assert_eq!(second_tag, Some("t2".to_owned()));
+    let (second_tag, stream_rest) = runtime.block_on(stream_rest.into_future());
+    assert_eq!(second_tag.unwrap().unwrap(), "t2".to_owned());
 
-    let (end, _) = runtime.block_on(stream_rest.into_future()).ok().unwrap();
-    assert_eq!(end, None);
+    let (end, _) = runtime.block_on(stream_rest.into_future());
+    assert!(end.is_none());
```
steveeJ: same as above

Cargo.toml
schrieveslaach: Sorry, the question is unclear. 

I wanted to ask, if I should bump the version of `dkregistry` itself. For example: 

```toml
[package]
name = "dkregistry"
version = "0.4.0-alpha.0"
authors = ["Luca Bruno <lucab@debian.org>", "Stefan Junker <sjunker@redhat.com>"]
license = "MIT/Apache-2.0"
```

lucab: Ah, yes please. `0.4.0-alpha.0` would be perfect.

tests/mock/api_version.rs
schrieveslaach: Are you looking for [`block_on`](https://docs.rs/futures/0.3.1/futures/executor/fn.block_on.html)?

Cargo.toml
schrieveslaach: I think and tested that dev-dependencies won't be exposed to users of this crate. If I try to use following code in a rust application, the compiler tells me that it cannot find the code that is only available behind the feature `macros`.

```rust
#[tokio::main]
async fn main() -> Result<(), Error> {
    // …
}
```

```
|
| #[tokio::main]
|          ^^^^ could not find `main` in `tokio`
```

tests/mock/catalog.rs
schrieveslaach: Unfortunately, I had to make this due to a limitation of [`error_chain`](https://docs.rs/error-chain/0.12.1/error_chain/):

> Because the error type contains std::error::Error+Send + 'static objects, it can't implement PartialEq for easy comparisons.

The original code results in a compiler error:

```
|     assert_eq!(page2, None);
|     ^^^^^^^^^^^^^^^^^^^^^^^^
|     |
|     std::option::Option<std::result::Result<std::string::String, mock::api_version::dkregistry::errors::Error>>
|     std::option::Option<_>
```

What do you propose to improve the code?

src/lib.rs
schrieveslaach: done

src/errors.rs
schrieveslaach: done

src/v2/catalog.rs
schrieveslaach: I don't have any experience with it. Do you mind to clean this up in another PR?

steveeJ: That's fine by me.

tests/mock/api_version.rs
steveeJ: Yes, that'll do it!

src/v2/catalog.rs
steveeJ: After taking another look at the code and spotting the `unsafe` which is used within this function, I suggest the following:

```suggestion
    pub async fn get_catalog(&self, paginate: Option<u32>) -> Result<String> {
```

schrieveslaach: `Result<String>` seems a bit limited too me because as a user I would have to parse the string. I think the method should return `Result<Vec<String>>`. What do you think?

steveeJ: `Result<String>` matches the previous ` Box<dyn futures::Stream<Item = String, Error = Error>>`. If we return a `Vec<String>` here, the caller will collect into a `Vec<Vec<String>>` which then needs to be flattened to `Vec<String>`. It would be more convenient for the consumer to collect into that directly. #140 implements my suggestion.

schrieveslaach: Okay. I wasn't aware that you can use futures for streams. But it makes totally sense.

schrieveslaach:
@steveeJ, how do you want to proceed with this PR? Do you want me to squash the commits and you finished the rest in #140?

steveeJ:
@schrieveslaach 

> Do you want me to squash the commits and you finished the rest in #140?

That sounds good to me. I'll rebase on top of your squashed branch, then Github will automatically detect this one as merged. I'd appreciate if you could help review #140 as well!!

schrieveslaach:
@steveeJ, I squashed the commits. I will have a look at #140 in the next couple of days.

steveeJ:
> @steveeJ, I squashed the commits. I will have a look at #140 in the next couple of days.

Ah, thanks for that, but could I ask you to split the commtis which touch the _Cargo.*_ files from the code changes, please? We have tried to separate these changes in case code changes need to be reverted without changing the dependencies at one point.

steveeJ:
I took the liberty of helping you out with splitting up the commits while preserving your authorship as best as I could. Please take a look at the commit history of #140 and let me know if that's fine for you. I suggest we either close this PR and focus on #140, but I'll leave it up to you.

schrieveslaach:
@steveeJ, that's fine for me. :+1:

PR #140, opened 2020-01-14
bump futures and friends and use async/await where possible

steveeJ:
This is based on #138 (thanks to @schrieveslaach!).

I'm happy to squash the new commits into the one's made by @schrieveslaach if we reach consensus on doing so.


tests/mock/tags.rs
```diff
@@ -73,16 +74,16 @@ fn test_tags_paginate() {
...
 
-    let (second_tag, stream_rest) = runtime.block_on(stream_rest.into_future()).ok().unwrap();
-    assert_eq!(second_tag, Some("t2".to_owned()));
+    let (second_tag, stream_rest) = runtime.block_on(stream_rest.into_future());
+    assert_eq!(second_tag.unwrap().unwrap(), "t2".to_owned());
 
-    let (end, _) = runtime.block_on(stream_rest.into_future()).ok().unwrap();
-    assert_eq!(end, None);
+    let (end, _) = runtime.block_on(stream_rest.into_future());
+    assert!(end.is_none());
```
steveeJ: FIXME: don't swallow output

tests/mock/catalog.rs
```diff
@@ -68,17 +72,17 @@ fn test_catalog_paginate() {
...
-    let (page1, next) = runtime.block_on(next.into_future()).ok().unwrap();
-    assert_eq!(page1, Some("r1/i1".to_owned()));
+    let (page1, next) = runtime.block_on(next.into_future());
+    assert_eq!(page1.unwrap().unwrap(), "r1/i1".to_owned());
 
-    let (page2, next) = runtime.block_on(next.into_future()).ok().unwrap();
+    let (page2, next) = runtime.block_on(next.into_future());
     // TODO(lucab): implement pagination
-    assert_eq!(page2, None);
+    assert!(page2.is_none());
```
steveeJ: FIXME: don't swallow output

tests/mock/catalog.rs
```diff
@@ -68,17 +72,17 @@ fn test_catalog_paginate() {
...
 
-    let (end, _) = runtime.block_on(next.into_future()).ok().unwrap();
-    assert_eq!(end, None);
+    let (end, _) = runtime.block_on(next.into_future());
+    assert!(end.is_none());
```
steveeJ: FIXME: don't swallow output

src/errors.rs
```diff
@@ -4,7 +4,6 @@ use base64;
 use http;
 use regex;
 use reqwest;
-use serde_json;
```
lucab: I think we can simply drop all the `use` statements in here.

src/v2/auth.rs
```diff
@@ -19,48 +16,43 @@ impl TokenAuth {
...
 impl Client {
-    fn get_token_provider(&self) -> FutureString {
+    async fn get_token_provider(&self) -> Result<String> {
         let url = {
             let ep = format!("{}/v2/", self.base_url.clone(),);
             match reqwest::Url::parse(&ep) {
                 Ok(url) => url,
                 Err(e) => {
-                    return Box::new(future::err::<_, _>(Error::from(format!(
+                    return Err(Error::from(format!(
```
lucab: Would a `bail!()` work here?

src/v2/auth.rs
```diff
@@ -215,6 +187,25 @@ fn parse_hdr_bearer(input: &str) -> Result<(String, Option<&str>)> {
...
+    if client.is_auth(None).await? {
+        return Ok(client);
+    }
+
+    let token = client.login(&[login_scope.as_str()]).await?;
+
+    if !client.is_auth(Some(token.token())).await? {
+        Err("login failed".into())
+    } else {
+        println!("logged in!");
```
lucab: Leftover or `log::trace!()`?

src/v2/auth.rs
```diff
@@ -215,6 +187,25 @@ fn parse_hdr_bearer(input: &str) -> Result<(String, Option<&str>)> {
     Ok((auth_ep, service))
 }
 
+pub async fn authenticate_client(mut client: Client, login_scope: String) -> Result<Client> {
```
lucab: Can we move this to a proper method of `&mut Client`?

src/v2/catalog.rs
```diff
@@ -1,60 +1,61 @@
...
-            let suffix = if let Some(n) = paginate {
-                format!("?n={}", n)
-            } else {
-                "".to_string()
+    pub fn get_catalog<'a, 'b: 'a>(
+        &'b self,
+        paginate: Option<u32>,
+    ) -> impl Stream<Item = Result<String>> + 'a {
+        try_stream! {
+            let url = {
```
lucab: Would it be possible to move this "once-init" logic out of the `try_stream!()` block? It doesn't seem to affect/changt with stream iteration, so having it outside of the macro block would improve readability.

src/v2/catalog.rs
```diff
@@ -1,60 +1,61 @@
...
-                })
-            })
-            .and_then(|body| -> Result<Catalog> {
-                serde_json::from_slice(&body).map_err(|e| e.into())
-            })
-            .map(|cat| futures::stream::iter_ok(cat.repositories.into_iter()))
-            .flatten_stream();
-        Box::new(fres)
+async fn fetch_catalog(req: RequestBuilder) -> Result<Catalog> {
+    match req.send().await {
```
lucab: I think a few indent-levels in this block logic can be dropped by immediately `?`ing the intermediate results. That should make the happy-path logic more obvious to follow.

src/v2/mod.rs
```diff
@@ -65,74 +59,52 @@ pub struct Client {
...
-        self.is_v2_supported()
-            .map(move |ok| (ok, self))
-            .and_then(|(ok, client)| {
-                if !ok {
-                    bail!("remote server does not support docker-registry v2 API")
-                } else {
-                    Ok(client)
-                }
-            })
+    pub async fn ensure_v2_registry(self) -> Result<Self> {
```
lucab: I think this can be changed to a simpler `&self` and `-> Result<()>`.

tests/mock/catalog.rs
```diff
@@ -29,7 +29,11 @@ fn test_catalog_simple() {
 
     let futcheck = dclient.get_catalog(None);
 
-    let res = runtime.block_on(futcheck.collect()).unwrap();
+    let (res, _errors): (Vec<_>, Vec<_>) = runtime
+        .block_on(futcheck.collect::<Vec<_>>())
+        .into_iter()
+        .partition(Result::is_ok);
+    let res: Vec<_> = res.into_iter().map(Result::unwrap).collect();
```
lucab: This should likely also assert that `errors.len() == 0`.

tests/mock/tags.rs
```diff
@@ -31,8 +31,9 @@ fn test_tags_simple() {
 
     let futcheck = dclient.get_tags(name, None);
 
-    let res = runtime.block_on(futcheck.collect()).unwrap();
-    assert_eq!(res, vec!["t1", "t2"]);
+    let res = runtime.block_on(futcheck.collect::<Vec<_>>());
+    assert_eq!(res.get(0).unwrap().as_ref().unwrap(), &String::from("t1"));
```
lucab: Can we unpack the steps to avoid the two chained `unwrap`s in a single line? That would improve line-referencing in case of test failures.

examples/tags.rs
```diff
@@ -58,17 +57,18 @@ fn run(
...
-                println!("{:?}", tag);
-            }
-            Ok(())
-        });
+    let dclient = common::authenticate_client(client, login_scope).await?;
 
-    match runtime.block_on(futures) {
-        Ok(_) => Ok(()),
-        Err(e) => Err(Box::new(e)),
+    let (tags, _errors): (Vec<_>, Vec<_>) = dclient
```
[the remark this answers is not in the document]
steveeJ: I assume you also mean to abort the iteration on error?

src/v2/catalog.rs
steveeJ: The `?` operator behaves differently inside the macro block. If I simply move the code outside, the compiler says:

```
error[E0277]: the `?` operator can only be used in  that returns `Result` or `Option` (or another type that implements `std::ops::Try`)
  --> src/v2/catalog.rs:26:13
   |
26 | /             reqwest::Url::parse(&ep).map_err(|e| {
27 | |                 Error::from(format!("failed to parse url from string '{}': {}", ep, e))
28 | |             })?
   | |_______________^ cannot use the `?` operator in  that returns `async_stream::async_stream::AsyncStream<std::result::Result<std::string::String, errors::Error>, impl core::future::future::Future>`
   |
   = help: the trait `std::ops::Try` is not implemented for `async_stream::async_stream::AsyncStream<std::result::Result<std::string::String, errors::Error>, impl core::future::future::Future>`
   = note: required by `std::ops::Try::from_error`
```

I might be able to work around this by manually constructing the stream error value, but I'm not sure if it's worth it. WDYT?

src/v2/auth.rs
steveeJ: The reason why I wrote this to consume the client is that it's easier in an asynchronous context. I can try again now to change that.

examples/tags.rs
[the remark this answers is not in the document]
lucab: Yes, I think it makes sense to mimic old-style stream behavior and stop+return the first error we encounter.

src/v2/catalog.rs
lucab: Not worth the complexity, I guess. However you can maybe still move most of it out of the macro block, and only keep the `?` inside. Assuming the resulting thing is not uglier. Your call.

src/v2/auth.rs
lucab: Ack, I'm also fine keeping it then.

You can drop the `mut` from the input parameters and re-bind internally. That would make consumers happier.

src/v2/mod.rs
steveeJ: Unless you feel strongly about this I'd rather not change the signature like that, as it'll brake consumers who might use this in a chained fashion.

tests/mock/catalog.rs
steveeJ: In that case we can probably just omit the partitioning.

src/v2/auth.rs
steveeJ: I don't think this `mut` matters to the consumer since the ownership is moved into the function. I did move it to be a method of `Client`, but other than that I left it unchanged.

steveeJ:
@lucab thanks a lot for the thorough review. I addressed all the comments. Please take another look.

lucab:
No further remarks from my side. I'll wait for the squash+rebase at https://github.com/camallo/dkregistry-rs/pull/138#issuecomment-576343665 to happen and then do the final stamp on this.

src/v2/catalog.rs
```diff
@@ -1,60 +1,53 @@
...
 struct Catalog {
     pub repositories: Vec<String>,
 }
 
 impl v2::Client {
-    pub fn get_catalog(&self, paginate: Option<u32>) -> StreamCatalog {
+    pub fn get_catalog<'a, 'b: 'a>(
+        &'b self,
+        paginate: Option<u32>,
+    ) -> impl Stream<Item = Result<String>> + 'a {
```
schrieveslaach: :+1:

tests/net/quay_io/mod.rs
```diff
@@ -70,29 +94,13 @@ fn test_quayio_insecure() {
     assert_eq!(res, true);
 }
 
+#[cfg(feature = "test-net-private")]
 #[test]
```
schrieveslaach: What about [`#[tokio::test]`](https://docs.rs/tokio/0.2.11/tokio/attr.test.html)? Then you wouldn't have to create a `Runtime` in each test.

steveeJ:
Thanks for the review @schrieveslaach!

> The tests could be made simpler by using `tokio::test` but that should not block the PR.

I gave it a quick try and would have given this more thought than I want to for the scope of this PR, when refactoring the tests which test the `impl Stream` functions in the mock tests. I'll be happy to do a follow-up PR with that refactor. I want to give you and @lucab another chance to express any strong feelings about blocking the PR on this, otherwise I think we're good to merge it.

schrieveslaach:
A follow-up PR sounds good to me.

lucab:
LGTM, thanks both of you!

lucab:
@schrieveslaach @steveeJ does any of you need this to be part of a release in the short term, or can we keep going and lazily tag one later on?

schrieveslaach:
Would be nice if I could use a release(-candidate) in aixigo/PREvant#28 but you don't have to hurry.

PR #148, opened 2020-04-01
v2/manifest: add a function to fetch content-digest for a tag

vrutkovs:
This is a lightweight alternative for get_manifest_and_ref, which
doesn't fetch the whole manifest. It can be used to ensure tag didn't
change content-digest.

This is a precondition to cache manifests in Cincinnati if the tag didn't change content digest.

Ref: https://issues.redhat.com/browse/OTA-167

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -32,45 +32,9 @@ impl Client {
         name: &str,
         reference: &str,
     ) -> Result<(Manifest, Option<String>)> {
-        let url = {
-            let ep = format!(
-                "{}/v2/{}/manifests/{}",
-                self.base_url.clone(),
-                name,
-                reference
-            );
-            match reqwest::Url::parse(&ep) {
-                Ok(url) => url,
-                Err(e) => {
-                    return Err(Error::from(format!(
-                        "failed to parse url from string '{}': {}",
-                        ep, e
-                    )));
-                }
-            }
-        };
+        let url = self.build_url(name, reference)?;
 
-        let accept_headers = header::HeaderMap::from_iter(
-            [
-                // accept header types and their q value, as documented in
-                // https://tools.ietf.org/html/rfc7231#section-5.3.2
-                (mediatypes::MediaTypes::ManifestV2S2, 0.5),
-                (mediatypes::MediaTypes::ManifestV2S1Signed, 0.4),
-                // TODO(steveeJ): uncomment this when all the Manifest methods work for it
-                // mediatypes::MediaTypes::ManifestList,
-            ]
-            .iter()
-            .filter_map(|(ty, q)| {
-                match header::HeaderValue::from_str(&format!("{}; q={}", ty.to_string(), q)) {
-                    Ok(header_value) => Some((header::ACCEPT, header_value)),
-                    Err(e) => {
-                        let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                        error!("{}", msg);
-                        None
-                    }
-                }
-            }),
-        );
+        let accept_headers = build_accept_headers();
 
         let client_spare0 = self.clone();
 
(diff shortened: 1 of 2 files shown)


src/v2/manifest/mod.rs
```diff
@@ -141,6 +105,63 @@ impl Client {
         }
     }
 
+    async fn build_url(&self, name: &str, reference: &str) -> Result<Url> {
```
lucab, line 108: I think this doesn't need to be `async`.

src/v2/manifest/mod.rs
```diff
@@ -294,6 +315,30 @@ fn evaluate_media_type(
     }
 }
 
+async fn build_accept_headers() -> header::HeaderMap {
```
lucab, line 318: I think this doesn't need to be `async`.

src/v2/manifest/mod.rs
```diff
@@ -141,6 +105,63 @@ impl Client {
...
+            Err(e) => {
+                return Err(Error::from(format!(
+                    "failed to parse url from string '{}': {}",
+                    ep, e
+                )));
+            }
+        }
+    }
+
+    /// Fetch content digest for a particular tag
```
lucab, line 126: Missing trailing dot.

src/v2/manifest/mod.rs
```diff
@@ -141,6 +105,63 @@ impl Client {
...
+    pub async fn get_manifestref(&self, name: &str, reference: &str) -> Result<Option<String>> {
+        let url = self.build_url(name, reference).await?;
+
+        let accept_headers = build_accept_headers().await;
+
+        let res = self
+            .build_reqwest(
+                reqwest::Client::new()
+                    .head(url.clone())
```
lucab, line 135: If this `url` is not used further on (it doesn't seem so), I think you don't need to `clone()` it here.

src/v2/manifest/mod.rs
```diff
@@ -141,6 +105,63 @@ impl Client {
...
+            )
+            .send()
+            .await?;
+
+        let status = res.status();
+        trace!("HEAD '{}' status: {:?}", res.url(), status);
+
+        match status {
+            StatusCode::OK => {}
+            _ => return Err(format!("GET {}: wrong HTTP status '{}'", res.url(), status).into()),
```
lucab, line 146: s/GET/HEAD/

src/v2/manifest/mod.rs
```diff
@@ -141,6 +105,63 @@ impl Client {
...
+        let ep = format!(
+            "{}/v2/{}/manifests/{}",
+            self.base_url.clone(),
+            name,
+            reference
+        );
+        match reqwest::Url::parse(&ep) {
+            Ok(url) => Ok(url),
+            Err(e) => {
+                return Err(Error::from(format!(
```
lucab, line 118: If you want to make this a bit more compact, I think it is simply a `.map_err(|e| format!(...).into())`

src/v2/manifest/mod.rs
vrutkovs: It does, otherwise `get_manifest_and_ref` call explodes, saying that Result doesn't implement Future

lucab: Sorry, my comment was not explicit enough.

You can drop the `async` here and at the same time also drop the matching `.await` when you call `build_url`. That is, `fn build_url` coupled with `self.build_url(x, y)?`.

vrutkovs: Oh, I trusted RLS way too much as it had a hiccup and threw and error initially. It does work now, pushed. Thanks!

lucab:
Code looks good! CI for the PR is green, but for some reason the notification had an hiccup.

Can you please add some integration tests at https://github.com/camallo/dkregistry-rs/tree/master/tests/net?
If this works with an anonymous client, we can cover all of quay + dockerhub + GCR. If it requires an authenticated user, the `test-net-private` feature is wired in Travis so that it can login to quay.io.

vrutkovs:
Hmm, this got a bit complicated.

I didn't add tests to unauth registries as we can't hardcode sha256 for the repos we don't control. Should I pick `coreos/alpine-sh` and hope it won't change?

In quay tests I don't have access to quay.io/steveej/cincinnati-test:0.0.1 and it appears Travis won't run these tests in my namespace, so I can't fish it out of the output :)

tests/net/quay_io/mod.rs
```diff
@@ -240,6 +240,32 @@ fn test_quayio_has_no_manifest() {
...
+    let actual = runtime.block_on(fut_actual).unwrap();
+    assert_eq!(actual, "");
+}
+
+#[cfg(feature = "test-net-private")]
+#[test]
+fn test_quayio_auth_manifestref() {
+    let image = "steveej/cincinnati-test";
+    let tag = "0.0.1";
+    let expected = String::from("FIXME");
```
lucab, line 260: `s/FIXME/sha256:cc1f79c6a6fc92982a10ced91bddeefb8fbd037a01ae106a64d0a7e79d0e4813/`

tests/net/quay_io/mod.rs
```diff
@@ -240,6 +240,32 @@ fn test_quayio_has_no_manifest() {
...
 }
 
+#[test]
+fn test_quayio_auth_manifestref_missing() {
+    let image = "steveej/cincinnati-test";
+    let tag = "no-such-tag";
+
+    let login_scope = format!("repository:{}:pull", image);
+    let (mut runtime, dclient) = common_init(Some(&login_scope)).unwrap();
+    let fut_actual = async { dclient.get_manifestref(image, tag).await.unwrap() };
```
lucab, line 250: This actually returns an `Err` as the tag does not exist. I think you wanted to `.is_err()` somewhere around it.

lucab:
@vrutkovs I'm fine with just going for a test on the private image we control, I've left the ref in a comment.

(I've no idea what's up with GitHub, I see the incoming webhook from Travis but the status is never updated)

lucab:
@vrutkovs I had to uninstall and redo the whole Travis setup, but now it seems to work. Tests are failing, Travis output has the expected values.

vrutkovs:
Done, had to remove a part of the test which verifies the error message - but I don't think its really necessary

PR #152, opened 2020-05-15
Reuse reqwest::Client

vrutkovs:
This commit ensures reqwest client is created once and reused across calls


src/v2/config.rs
```diff
@@ -81,12 +81,15 @@ impl Config {
                 p.unwrap_or_else(|| "".into()),
             )),
         };
+        let client = reqwest::ClientBuilder::new().build().unwrap();
```
lucab, line 84: I think this `.unwrap()` is a development leftover.

PR #156, opened 2020-06-14
Use thiserror for error handling

MikailBag:
# Motivation
 - `error-chain` generates Error type which is not `Sync`, so it is hard to use with many other libraries.
 - Libraries are encouraged to provide structured error details instead of just formatted message.
# Changes
 - `errors::Error` type is now plain enum, using `thiserror` for `Display` and `From` impls.
 - It is very large, but I don't think it is significant drawback.
 - Some functions return custom Error type (which is more convenient to use), but such types can be converted to `errors::Error`.
 - Unfortunately new approach does not play well with backtraces. I think this can be fixed if needed.
 - In several places I replaced `?` with `expect` because there should not be any errors in fact.

### DIFF
--- a/src/mediatypes.rs
+++ b/src/mediatypes.rs
@@ -1,6 +1,6 @@
 //! Media-types for API objects.
 
-use crate::errors::{Error, Result};
+use crate::errors::{Result};
 use mime;
 use strum::EnumProperty;
 
@@ -62,13 +62,13 @@ impl MediaTypes {
                     }
                     ("vnd.docker.image.rootfs.diff.tar.gzip", _) => Ok(MediaTypes::ImageLayerTgz),
                     ("vnd.docker.container.image.v1", "json") => Ok(MediaTypes::ContainerConfigV1),
-                    _ => bail!("unknown mediatype {:?}", mtype),
+                    _ => return Err(crate::Error::UnknownMimeType(mtype.clone())),
                 }
             }
-            _ => bail!("unknown mediatype {:?}", mtype),
+            _ => return Err(crate::Error::UnknownMimeType(mtype.clone())),
         }
     }
-    pub fn to_mime(&self) -> Result<mime::Mime> {
+    pub fn to_mime(&self) -> mime::Mime {
         match self {
             &MediaTypes::ApplicationJson => Ok(mime::APPLICATION_JSON),
             ref m => {
@@ -79,6 +79,6 @@ impl MediaTypes {
                 }
             }
         }
-        .map_err(|e| Error::from(e.to_string()))
+        .expect("to_mime should be always successful")
     }
 }
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -1,7 +1,6 @@
 use crate::errors::{Error, Result};
 use crate::v2::*;
 use reqwest::{header::HeaderValue, RequestBuilder, StatusCode, Url};
-use std::iter::FromIterator;
 
 /// Represents all supported authentication schemes and is stored by `Client`.
 #[derive(Debug, Clone)]
@@ -41,12 +40,7 @@ impl BearerAuth {
         let auth_ep = bearer_header_content.auth_ep(scopes);
         trace!("authenticate: token endpoint: {}", auth_ep);
 
-        let url = reqwest::Url::parse(&auth_ep).map_err(|e| {
-            Error::from(format!(
-                "failed to parse url from string '{}': {}",
-                auth_ep, e
-            ))
-        })?;
+        let url = reqwest::Url::parse(&auth_ep)?;
 
         let auth_req = {
             Client {
@@ -65,14 +59,13 @@ impl BearerAuth {
         let status = r.status();
         trace!("authenticate: got status {}", status);
         if status != StatusCode::OK {
-            bail!("authenticate: wrong HTTP status '{}'", status);
+            return Err(Error::UnexpectedHttpStatus(status));
         }
 
         let bearer_auth = r.json::<BearerAuth>().await?;
 
         match bearer_auth.token.as_str() {
-            "unauthenticated" => bail!("token is unauthenticated"),
-            "" => bail!("received an empty token"),
+            "unauthenticated" | "" => return Err(Error::InvalidAuthToken(bearer_auth.token)),
             _ => {}
         };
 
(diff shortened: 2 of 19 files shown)


lucab:
@steveeJ I think you were looking at refreshing error handling in other places too, do you maybe have some cycles to review  and land this one? I'm +1 on `thiserror` here.

steveeJ:
> I think you were looking at refreshing error handling in other places too, do you maybe have some cycles to review and land this one? I'm +1 on `thiserror` here.

Yes, I'll be happy to take a look at this. I migrated [away from `failure` on openshift/cincinnati recently](https://github.com/openshift/cincinnati/pull/286). That PR was less invasive than this one, although I do understand the motivation of adding more specific and structured errors.

From glancing at this it looks like it'll be in conflict with #159, so I'd prefer to land that first.

lucab:
@steveeJ ack, let's serialize these. I started going through #159 but didn't manage to finish, I'll hope to finish that on Monday.

MikailBag:
Thanks for positive feedback!
I will be happy to rebase onto master and fix merge conflicts if needed.

steveeJ:
@MikailBag #159 has landed. I'll review this PR when it's rebased!

MikailBag:
I have rebased onto master; tests pass.

src/mediatypes.rs
```diff
@@ -79,6 +79,6 @@ impl MediaTypes {
                 }
             }
         }
-        .map_err(|e| Error::from(e.to_string()))
+        .expect("to_mime should be always successful")
```
steveeJ, line 82: I don't follow why this cannot fail. I'd prefer to leave the `Result` in place here, and let the caller make use of the `?` operator which isn't expensive.

src/v2/auth.rs
```diff
@@ -337,7 +315,7 @@ mod tests {
         let header_value = HeaderValue::from_str(&format!(
             r#"Bearer realm="{}",service="{}",scope="{}""#,
             realm, service, scope
-        ))?;
+        )).expect("this statically known header value only contains ASCII chars so it is correct header value");
```
steveeJ, line 318: The variables _realm_, _service_, and _scope_ aren't statically known.
Also, my rule of thumb is that if it's a 3rd party library I don't rely on it and avoid panicing.

src/v2/auth.rs
```diff
@@ -365,7 +343,7 @@ mod tests {
     fn basic_realm_parses_correctly() -> Result<()> {
         let realm = "Registry realm";
 
-        let header_value = HeaderValue::from_str(&format!(r#"Basic realm="{}""#, realm))?;
+        let header_value = HeaderValue::from_str(&format!(r#"Basic realm="{}""#, realm)).unwrap();
```
steveeJ, line 346: My rule of thumb is that if it's a 3rd party library I don't rely on it and avoid panicing.

src/mediatypes.rs
```diff
@@ -62,13 +62,13 @@ impl MediaTypes {
...
-                    _ => bail!("unknown mediatype {:?}", mtype),
+                    _ => return Err(crate::Error::UnknownMimeType(mtype.clone())),
                 }
             }
-            _ => bail!("unknown mediatype {:?}", mtype),
+            _ => return Err(crate::Error::UnknownMimeType(mtype.clone())),
         }
     }
-    pub fn to_mime(&self) -> Result<mime::Mime> {
+    pub fn to_mime(&self) -> mime::Mime {
```
steveeJ, line 71:
```suggestion
    pub fn to_mime(&self) -> Result<mime::Mime> {
```

Please see the comment below

src/v2/content_digest.rs
```diff
@@ -1,36 +1,47 @@
-use crate::errors::Result;
 /// Implements types and methods for content verification
 use sha2::{self, Digest};
 
 /// ContentDigest stores a digest and its DigestAlgorithm
 #[derive(Clone, Debug, PartialEq)]
-pub(crate) struct ContentDigest {
+pub struct ContentDigest {
```
steveeJ, line 6: Does this need to be `pub`?

src/v2/content_digest.rs
```diff
@@ -1,36 +1,47 @@
...
     digest: String,
     algorithm: DigestAlgorithm,
 }
 
 /// DigestAlgorithm declares the supported algorithms
 #[derive(Display, Clone, Debug, PartialEq, EnumString)]
-enum DigestAlgorithm {
+pub enum DigestAlgorithm {
```
steveeJ, line 13: Does this need to be `pub`?

src/v2/content_digest.rs
```diff
@@ -1,36 +1,47 @@
...
+}
+
 impl ContentDigest {
     /// try_new attempts to parse the digest string and create a ContentDigest instance from it
     ///
     /// Success depends on
     /// - the string having a "algorithm:" prefix
     /// - the algorithm being supported by DigestAlgorithm
-    pub fn try_new(digest: String) -> Result<Self> {
+    pub fn try_new(digest: String) -> std::result::Result<Self, ContentDigestError> {
```
steveeJ, line 37: What's the advantage of declaring this specific error type in this module vs. adding the variants to the crate's error module?

MikailBag: If one wishes to use specifically ContestDigest, they can exhaustively match on ContestDigestError and handle all possible errors. If this function returns `crate::Error` instead, such a user would have to 
1) add wildcard check
2) write justyfing comment
3) carefully read through docs of `dkregistry::Error` (or source code) and decide for each variant if it is possible.
It is time-consuming process.

In short: IMHO, if some function returns E as error enum, and it contains variants that are never returned by this function, it is inconvenient for API user who would like to do exhaustive match.


That said, I'm ready to merge this domain-specific small enums into `crate::Error`, if you find that behavior more maintainable or readable,

src/v2/content_digest.rs
MikailBag: I changed visibility to `pub` because I included it into `ContestDigestError::Verify` variant details.
I *think* this can be useful for library user.
However, I see alternative option: we can include `format!`-ed digests into error instead. this way we can keep ContestDigest private.

src/v2/content_digest.rs
steveeJ: Thanks for elaborating, the reasoning makes sense to me! It would be nice if Rust allowed a convenient way to specify a subset of variants the of the error enum a function returns, but I can't come up with an easy way to do that. In this case I prefer the specific error type as implemented in this PR.

MikailBag:
@steveeJ rebased.

src/v2/auth.rs
steveeJ: Bump.

src/v2/auth.rs
steveeJ: Bump.

steveeJ (changes requested):
I spotted two unresolved comment threads and bumped them. I strongly prefer not to panic if we can avoid it.

MikailBag:
Correct me if I'm wrong, but now I can see that the two mentioned lines are in tests (sorry that I have not noticed it earlier). Usually it is OK for test code to panic.
I can change `expect` to `?` in that code, but I'd like to get some confirmation from you.

src/v2/auth.rs
steveeJ: Ah, I see now that they are statically known here by expanding the previous lines. Also as you noted in your comment, this is in a test so panicking is appropriate.

steveeJ (approved):
> Correct me if I'm wrong, but now I can see that the two mentioned lines are in tests (sorry that I have not noticed it earlier). Usually it is OK for test code to panic.

I overlooked the fact that it's in test as well. I agree that we can stick with it there.

Thank you for your work on this!

PR #157, opened 2020-06-19
tests: improve private feature gate handling; examples: add script to run all and make idempotent

steveeJ:
This makes test and example execution easier in local development environments, and gives us the option to easily run all examples on CI.

### DIFF
--- a/examples/image.rs
+++ b/examples/image.rs
@@ -5,13 +5,14 @@ extern crate tokio;
 
 use dkregistry::render;
 use futures::future::try_join_all;
+use std::path::Path;
 use std::result::Result;
 use std::{boxed, env, error, fs, io};
 
 mod common;
 
 #[tokio::main]
-async fn main() {
+async fn main() -> Result<(), boxed::Box<dyn error::Error>> {
     let registry = match std::env::args().nth(1) {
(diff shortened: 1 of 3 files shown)


examples/image.rs
```diff
@@ -99,7 +99,14 @@ async fn run(
     let path = &format!("{}:{}", &image, &version).replace("/", "_");
     let path = std::path::Path::new(&path);
     if path.exists() {
-        return Err(format!("path {:?} already exists, exiting", &path).into());
+        let msg = format!("path {:?} already exists", &path);
```
lucab, line 102: What about moving all this path cleaning/creation outside of the `async` block into its own function? It is blocking logic anyway and it is a bit distracting when readying through the registry-fetching logic. I think it can be safely performed upfront.

examples/run-all.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
```
lucab, line 1: Where you planning to wire this to Travis too maybe?

steveeJ: Yes, but not right away as I'm not sure this branch has all the examples working.

PR #158, opened 2020-06-19
v2/client: add accept_invalid_certs option

steveeJ:
This allows to ignore invalid TLS certificates which can be useful in
test environments.

### DIFF
--- a/src/v2/config.rs
+++ b/src/v2/config.rs
@@ -8,6 +8,7 @@ pub struct Config {
     user_agent: Option<String>,
     username: Option<String>,
     password: Option<String>,
+    accept_invalid_certs: bool,
 }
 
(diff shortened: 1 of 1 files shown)


lucab:
LGTM, but what do you think about mirroring the underlying name (i.e. keeping the `danger` prefix)?

steveeJ:
> what do you think about mirroring the underlying name (i.e. keeping the `danger` prefix)?

A bit too dramatic for my taste, hence I skipped it. I won't oppose adding it if you prefer that though.

lucab:
But the audience loves drama! Jokes aside, it was only to keep consistency with the `reqwest` client. We can freely take a different route. LGTM.

PR #159, opened 2020-06-19
v2/auth: add Basic authentication and refactor

steveeJ:
This adds Basic authentication support.

Along the way, the auth module has been refactored with new enums and
structs to model the supported authentication mechanisms and make the
response header parsing more robust.

***Breaking API changes***

* The `Client::login` and `Client::authenticate` methods have been
  merged into the latter.
* `Client::is_auth` now performs a pure authentication test without
  attempting to authenticate, thus it no longer takes a token.


src/v2/auth.rs
```diff
@@ -1,23 +1,165 @@
...
+            Auth::Basic(basic_auth) => {
+                request_builder.basic_auth(basic_auth.user.clone(), basic_auth.password.clone())
+            }
+        }
+    }
+}
 
-#[derive(Debug, Default, Deserialize, Serialize)]
-pub struct TokenAuth {
+/// Used for Bearer HTTP Authentication
```
steveeJ, line 25: Missing `.`

src/v2/auth.rs
```diff
@@ -1,23 +1,165 @@
...
     token: String,
     expires_in: Option<u32>,
     issued_at: Option<String>,
     refresh_token: Option<String>,
 }
 
-impl TokenAuth {
-    pub fn token(&self) -> &str {
-        self.token.as_str()
+/// Used for Basic HTTP Authentication
```
steveeJ, line 34: missing `.`

src/v2/auth.rs
```diff
@@ -1,23 +1,165 @@
...
+        let method = captures
+            .get(0)
+            .ok_or_else(|| {
+                Error::from(format!("regex '{}' didn't match '{}'", re.as_str(), header))
+            })?
+            .name("method")
+            .ok_or_else(|| Error::from(format!("method not found in {}", header)))?
+            .as_str();
+
+        let serialized_content = serde_json::Value::Object(serde_json::Map::from_iter(
```
steveeJ, line 80: This is somewhat inconvenient. I'm thinking of assembling a JSON string using plain string formatting and skipping the `Value::Object` route.

src/v2/auth.rs
```diff
@@ -1,23 +1,165 @@
...
+        let mut unsupported_keys = std::collections::HashSet::new();
+        let content: WwwAuthenticateHeaderContent = serde_ignored::deserialize(
+            &mut serde_json::Deserializer::from_str(&serialized_content),
+            |path| {
+                unsupported_keys.insert(path.to_string());
+            },
+        )?;
+
+        if !unsupported_keys.is_empty() {
+            warn!("unsupported keys remaining {:#?}", unsupported_keys);
```
steveeJ, line 112: Previously we threw an error in this case. I decided to change this because I think that as long as the parsing was successful, we may just work with the result and not be overly strict.

src/v2/auth.rs
```diff
@@ -35,91 +177,111 @@ impl Client {
...
+            .map(ToOwned::to_owned)
     }
 
-    /// Set the token to be used for further registry requests.
-    pub fn set_token(&mut self, token: Option<&str>) -> &Self {
-        if let Some(ref t) = token {
-            self.token = Some(t.to_string());
-        }
+    /// Set the authentication information to be used for further registry requests.
+    pub fn set_auth(&mut self, auth: Auth) -> &Self {
```
steveeJ, line 192: This is currently unused and can probably be removed.

src/v2/auth.rs
```diff
@@ -35,91 +177,111 @@ impl Client {
...
-        let mask_start = std::cmp::min(1, chars_count - 1);
-        let mask_end = std::cmp::max(chars_count - 1, 1);
-        t.replace_range(mask_start..mask_end, &"*".repeat(mask_end - mask_start));
+        if !self.is_auth().await? {
+            self.auth = None;
+            bail!("login failed")
+        }
 
-        trace!("login: got token: {:?}", t);
+        trace!("authenticate: login to succeeded");
```
steveeJ, line 276: typo: superfluous `to`

src/v2/auth.rs
steveeJ: Added https://github.com/camallo/dkregistry-rs/pull/159/commits/68da0dc2f657c2566ebbba8f927268b4c11e50e3 on top which implements the plain string route. I like the aspect that we don't need the intermediate `Value` and I can't spot any weakness in the approach. I'd appreciate thoughts here.

src/v2/auth.rs
steveeJ: Added https://github.com/camallo/dkregistry-rs/pull/159/commits/68da0dc2f657c2566ebbba8f927268b4c11e50e3 on top which implements the plain string route. I like the aspect that we don't need the intermediate `Value` and I can't spot any weakness in the approach. I'd appreciate thoughts here.

src/v2/auth.rs
```diff
@@ -1,23 +1,161 @@
...
+                Error::from(format!("regex '{}' didn't match '{}'", re.as_str(), header))
+            })?
+            .name("method")
+            .ok_or_else(|| Error::from(format!("method not found in {}", header)))?
+            .as_str();
+
+        let serialized_content = format!(
+            r#"{{ "{}": {{ {} }} }}"#,
+            method,
+            captures
```
lucab, line 83: Nit: it would be a bit more readable having this iter-collection logic outside of the `format!()`.

src/v2/auth.rs
lucab: I don't have strong opinions in either directions right now. If we stick to a warning log, we should add a bit of context like "skipping unrecognized keys in authentication header: {:#?}".

src/v2/auth.rs
```diff
@@ -1,23 +1,161 @@
...
+            .service
+            .as_ref()
+            .map(|sv| format!("?service={}", sv))
+            .unwrap_or_default();
+
+        let scope = scopes.iter().enumerate().fold(
+            if scopes.is_empty() {
+                ""
+            } else if service.is_empty() {
+                "?"
```
lucab, line 135: I have the feeling this possibly does not belong to the folding accumulator. I would have expected this one on its own `separator`, outside of the iterator and directly plugged into the format string. What do you think?

src/v2/auth.rs
```diff
@@ -1,23 +1,161 @@
...
+    realm: String,
+}
+
 impl Client {
-    async fn get_token_provider(&self) -> Result<String> {
+    /// Make a request and return the response's www authentication header.
+    async fn get_www_authentication_header(&self) -> Result<HeaderValue> {
         let url = {
             let ep = format!("{}/v2/", self.base_url.clone(),);
             match reqwest::Url::parse(&ep) {
```
lucab, line 161: If you don't mind while re-arranging this function, can this be turned into a `.map_err()?` maybe?

src/v2/auth.rs
```diff
@@ -35,91 +173,105 @@ impl Client {
...
+            })
+            .map(ToOwned::to_owned)
+    }
 
-        let (mut auth_ep, service) = parse_hdr_bearer(chal.trim_start_matches("Bearer "))?;
+    /// Perform registry authentication and return the authenticated client.
+    ///
+    /// If Bearer authentication is used the returned client will be authorized for the requested scopes.
+    pub async fn authenticate(mut self, scopes: &[&str]) -> Result<Self> {
+        let credentials = if let Some(credentials) = self.credentials.clone() {
```
lucab, line 191: This looks like a `.cloned().ok_or_else(...)?`, right?

src/v2/auth.rs
```diff
@@ -35,91 +173,105 @@ impl Client {
...
-        self
-    }
+                let url = reqwest::Url::parse(&auth_ep).map_err(|e| {
+                    Error::from(format!(
+                        "failed to parse url from string '{}': {}",
+                        auth_ep, e
+                    ))
+                })?;
+
+                let auth_req = match self.credentials.clone() {
```
lucab, line 218: I think you already have a cloned `credentials` available here from the outer scope.

src/v2/auth.rs
```diff
@@ -35,91 +173,105 @@ impl Client {
...
-        if let Some(sv) = service {
-            auth_ep += &format!("?service={}", sv);
-            trace!("Service identity: {}", sv);
-        }
+        self.auth = None;
 
-        Ok(auth_ep)
-    }
+        let authentication_header = self.get_www_authentication_header().await?;
+        match WwwAuthenticateHeaderContent::from_www_authentication_header(authentication_header)? {
```
lucab, line 200: This logic is not very nice to follow. I would suggest:
 * doing the `self.auth` assignment in a single place, at the bottom of the function
 * move out the logic in this match statement, from the Bearer arm into its own async method attached to `BearerAuth`

src/v2/auth.rs
```diff
@@ -35,91 +173,105 @@ impl Client {
...
+        } else {
+            bail!("cannot authenticate without credentials");
+        };
 
-        trace!("Token provider: {}", auth_ep);
-        if let Some(sv) = service {
-            auth_ep += &format!("?service={}", sv);
-            trace!("Service identity: {}", sv);
-        }
+        self.auth = None;
```
lucab, line 197: I think is not needed. If it is actually needed, it's a good sign that we should re-arrange this logic because we have unwanted side-effects leaking into the authentication logic from previous state.

src/v2/auth.rs
```diff
@@ -35,91 +173,105 @@ impl Client {
...
+            }
         };
 
-        // mask the token before logging it
-        let chars_count = t.chars().count();
-        let mask_start = std::cmp::min(1, chars_count - 1);
-        let mask_end = std::cmp::max(chars_count - 1, 1);
-        t.replace_range(mask_start..mask_end, &"*".repeat(mask_end - mask_start));
+        if !self.is_auth().await? {
+            self.auth = None;
```
lucab, line 262: I think this does not have any effect, but see comments above regarding doing a single assignment if possible.

src/v2/auth.rs
steveeJ: @lucab any opinion on this one?

src/v2/auth.rs
steveeJ: On the second look I noticed that the match logic is covered earlier, so even better ;-)

src/v2/auth.rs
lucab: [sorry I didn't see this comment in review mode in GH]

The main problem I see with the logic in the fixup commit is when either `key` or `value` contains some characters that may need to be escaped in JSON-encoding (e.g. `"`). serde serialization takes care of that for you and is in general more robust.

But I do agree this code is a bit unwieldy. I think the source of the pain is the round-trip through JSON and serde. As a manual alternative, I think you could shove the key-value pairs into a map, drain the well-known values you expect in order to build the object, and warn if there are any leftover keys in the map.

src/v2/auth.rs
steveeJ: It's a relic from the earlier code which double-checks that the authentication was successful. I'm confident enough to remove it.

steveeJ:
Thanks for the review @lucab! I addressed all comments in a fixup commit. I'll think about https://github.com/camallo/dkregistry-rs/pull/159#discussion_r445542278 a bit more.

src/v2/auth.rs
steveeJ: I found a compromise by wrapping the strings in a `serde_json::Value::String`. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=721415923973d82b083b7e82f6d5e320). WDYT?

steveeJ: I pushed this suggestion into the fixup commit and am proposing it as the final solution. It doesn't have the escaping bug you detected with the previous revision, and it doesn't do the round trip. Hence I like 0ba493ed37ee2a5d71713664e2bcae76cd2ed3c6 and I'm proposing it as the final solution. PTAL and let me know if I should squash :wink:

src/v2/auth.rs
steveeJ: Ops, I removed this conditional altogether now as discussed in the stroked text at https://github.com/camallo/dkregistry-rs/pull/159#discussion_r445542963:

> It _[the `self.is_auth()` check]_ is a relic from the earlier code which double-checks that the authentication was successful. I'm confident enough to remove it. (removed because I thought this was on a different line)

 WDYT?

steveeJ:
> Using my limited knowledge of Rust and dkregistry-rs the code changes looks fine to me.

To add some context to this, @LalatenduMohanty and I went over the code and the changes together using screen-sharing for about two hours :wink:

src/v2/auth.rs
lucab: Yeah, it looks better than the initial version. Feel free to squash in.

src/v2/auth.rs
lucab: Yes, I don't remember adding this for any specific reason, I think it was mostly an overly abundant amount of safety check.

lucab:
Yes, it also looks good to me. Feel free to squash the fixup and self-merge.

PR #162, opened 2020-06-25
Workaround GCR bug

### LINKED ISSUE #161: Does not work with GCR with default settings
`get_manifest` fails, because GCR returns error 404.
Response body is 
```json
{"errors":[{"code":"MANIFEST_UNKNOWN","message":"Manifest with tag 'latest' has media type 'application/vnd.docker.distribution.manifest.v2+json', but client accepts 'application/vnd.docker.distribution.manifest.v2+json; q=0.5,application/vnd.docker.distribution.manifest.v1+prettyjws; charset=utf-8; q=0.4'."}]}
```
I am not sure if I right, but it *seems* like incorrect parsing of `Accept` HTTP header on GCR side.
What do you think?

MikailBag:
Resolves #161

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -34,7 +34,7 @@ impl Client {
     ) -> Result<(Manifest, Option<String>)> {
         let url = self.build_url(name, reference)?;
 
-        let accept_headers = build_accept_headers();
+        let accept_headers = build_accept_headers(&self.index);
 
         let client_spare0 = self.clone();
 
@@ -117,7 +117,7 @@ impl Client {
     pub async fn get_manifestref(&self, name: &str, reference: &str) -> Result<Option<String>> {
         let url = self.build_url(name, reference)?;
 
-        let accept_headers = build_accept_headers();
+        let accept_headers = build_accept_headers(&self.index);
 
         let res = self
             .build_reqwest(Method::HEAD, url)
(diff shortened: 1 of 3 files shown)


MikailBag:
Manual tests show that public image from GCR is now pulled successfully (I have not checked pulling a private image though).
If you'd like, I can try to add test that manifest can be successfully downloaded.

lucab:
@MikailBag thanks for the invesigation. We do have some integration tests that are running against GCR, but I think they are not hitting this (possibly because they do not exercise that path). Do you think it would be possible to expose #161 via Travis, and then fix it here?

MikailBag:
Do you mean adding test [here](https://github.com/camallo/dkregistry-rs/blob/master/tests/net/gcr_io/mod.rs)?

MikailBag:
@steveeJ ok, I will add `gcr.io` and `*.gcr.io` detection to this PR.

src/v2/manifest/mod.rs
```diff
@@ -286,7 +286,21 @@ fn evaluate_media_type(
     }
 }
 
-fn build_accept_headers() -> header::HeaderMap {
+fn build_accept_headers(registry: &str) -> header::HeaderMap {
+    if registry.ends_with("://gcr.io") || registry.ends_with(".gcr.io") {
```
steveeJ, line 290: I'm happy to learn why you chose these conditions.

I've got a hunch that we can just check if it ends with `gcr.io` here, as the TLD is very unlikely to be overriden in a local DNS system, and if so, they would probably know what they're doing.

MikailBag: well, I had following cases in mind:
"http://gcr.io" - true
"https://eu.gcr.io" - true
"http://my-super-gcr.io" - false

src/v2/manifest/mod.rs
```diff
@@ -286,28 +286,45 @@ fn evaluate_media_type(
...
-                    let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                    error!("{}", msg);
-                    None
+fn build_accept_headers(registry: &str) -> header::HeaderMap {
+    use strum::AsStaticRef;
+
+    // GCR incorrectly parses `q` parameters, so we use special Accept for it.
+    // Bug: https://issuetracker.google.com/issues/159827510.
+    // TODO: when bug is fixed, this workaround should be removed.
+    let no_q = registry.ends_with("://gcr.io") || registry.ends_with(".gcr.io");
```
steveeJ, line 295: The protocol [cannot be specified in a reference](https://github.com/docker/distribution/blob/0d3efadf0154c2b8a4e7b6621fff9809655cc580/reference/reference.go#L4), so we should never encounter `://` here. Did you actually encounter this somehow?

If I'm not mistaken, this should work:

```suggestion
    let no_q = registry == "gcr.io" || registry.ends_with(".gcr.io");
```

src/v2/manifest/mod.rs
```diff
@@ -286,28 +286,45 @@ fn evaluate_media_type(
...
-        .filter_map(|(ty, q)| {
-            match header::HeaderValue::from_str(&format!("{}; q={}", ty.to_string(), q)) {
-                Ok(header_value) => Some((header::ACCEPT, header_value)),
-                Err(e) => {
-                    let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                    error!("{}", msg);
-                    None
+fn build_accept_headers(registry: &str) -> header::HeaderMap {
+    use strum::AsStaticRef;
+
```
steveeJ, line 291: Sorry, this is a leftover from me. We don't actually need this.

```suggestion
```

src/v2/manifest/mod.rs
MikailBag: Well, I'm quite bad in naming :)
This functions in fact receives registry API endpoint, which [contains](https://github.com/camallo/dkregistry-rs/blob/d6dc42dc8e729fe6c5f8f4792e8b445f4c29eae4/src/v2/config.rs#L74) schema too.
I can either rename this parameter, or somehow change `Client` (for example it can store registry hostname and scheme separately).

src/v2/manifest/mod.rs
```diff
@@ -34,13 +34,13 @@ impl Client {
     ) -> Result<(Manifest, Option<String>)> {
         let url = self.build_url(name, reference)?;
 
-        let accept_headers = build_accept_headers();
-
         let client_spare0 = self.clone();
 
+        let headers = build_accept_headers(&self.base_url);
```
steveeJ, line 39:
```suggestion
        let headers = build_accept_headers(&self.index);
```

src/v2/manifest/mod.rs
steveeJ: I understand the confusion now and I added another suggestion to complement the existing one. I'd prefer to not deal with the schema here.

src/v2/manifest/mod.rs
MikailBag: :facepalm:

src/v2/manifest/mod.rs
```diff
@@ -286,28 +286,43 @@ fn evaluate_media_type(
...
-                Ok(header_value) => Some((header::ACCEPT, header_value)),
-                Err(e) => {
-                    let msg = format!("failed to parse HeaderValue from str: {}:", e);
-                    error!("{}", msg);
-                    None
+fn build_accept_headers(registry: &str) -> header::HeaderMap {
+    // GCR incorrectly parses `q` parameters, so we use special Accept for it.
+    // Bug: https://issuetracker.google.com/issues/159827510.
+    // TODO: when bug is fixed, this workaround should be removed.
+    let no_q = registry.ends_with("gcr.io") || registry.ends_with(".gcr.io");
```
steveeJ, line 293:
```suggestion
    let no_q = registry == "gcr.io" || registry.ends_with(".gcr.io");
```

The previous statement would evaluate to true for strings which you previously intended to exclude in https://github.com/camallo/dkregistry-rs/pull/162#discussion_r446432719, namely `my-super-gcr.io`.

MikailBag: Committed and squashed, thanks.

PR #175, opened 2020-09-09
fix: support 'charset=utf-8' parameter for application/json mimetype

edwardgeorge:
Seeing the charset param for application/json content-type present in the wild (in this case Azure CR), which was causing a strum error to be thrown when trying to construct the `MediaTypes` enum from the header.

Following the existing solution above in the enum this simply adds a second 'serialize' attribute param to the ApplicationJson variant.

It is noted that @steveeJ added a 'TODO' to find a generic way to handle these on the existing usage, but this is not addressed in this bugfix.

### DIFF
--- a/src/mediatypes.rs
+++ b/src/mediatypes.rs
@@ -18,7 +18,7 @@ pub enum MediaTypes {
         serialize = "application/vnd.docker.distribution.manifest.v1+prettyjws",
         to_string = "application/vnd.docker.distribution.manifest.v1+prettyjws",
 
-        // TODO(steveeJ) find a generic way to handle this form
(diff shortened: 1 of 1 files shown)


edwardgeorge:
Ah, i just noticed that this is a duplicate of #113 which I somehow missed. So I will close this and comment there.

edwardgeorge:
> Ah, i just noticed that this is a duplicate of #113 which I somehow missed. So I will close this and comment there.

reading the discussion on that PR, this looks to be the correct fix for the issue I was experiencing: https://github.com/camallo/dkregistry-rs/pull/176

PR #176, opened 2020-09-09
fix has_manifest for 404 responses - do not evaluate_media_type()

edwardgeorge:
If getting a 404 response when calling `has_manifest` then if the content-type returned for the json error body is `application/json; charset=utf-8` then the call to `evaluate_media_type` will fail with a strum error because it does not expect the `charset=utf-8` param to be valid here. see discussion in: https://github.com/camallo/dkregistry-rs/pull/113

This change follows the approach taken in other methods above which is to match on `status` before continuing to inspect the media type.

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -171,24 +171,26 @@ impl Client {
             .map_err(Error::from)?;
 
         let status = r.status();
-        let media_type = evaluate_media_type(r.headers().get(header::CONTENT_TYPE), &r.url())?;
 
         trace!(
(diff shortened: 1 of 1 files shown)


steveeJ (commented):
Thank you for this contribution @edwardgeorge!

I took the liberty to push some opinionated style changes directly to your fork as it seems easier to read for me that way. Please squash the commit into yours if you like it, otherwise feel free to drop it. Either way I'll merge the PR once you tell me to :wink:

edwardgeorge:
@steveeJ yep, those changes make sense. I have gone ahead and squashed them, feel free to merge.

PR #177, opened 2020-09-22
fix: correctly separate multiple scopes with '&'

edwardgeorge:
was generating invalid query strings with multiple scopes, eg:

```
.authenticate(&["repository:foo:pull", "repository:bar:pull"])
```
is serialised as:
```
?service=...&scope=repository:foo:pullscope=repository:bar:pull
```

note the missing '&' between pull and scope.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ dirs = "3.0"
 env_logger = "0.7"
 mockito = "0.27"
 spectral = "0.6"
+test-case = "1.0.0"
 tokio = { version = "0.2", features = ["macros"] }
 
 [features]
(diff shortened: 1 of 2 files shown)


lucab:
@edwardgeorge good catch! If you don't mind, can you please add unit tests for the following three cases:
 * no scopes
 * one scope
 * more than one scope

edwardgeorge:
@lucab sure, no problem.

edwardgeorge:
@lucab updated with three tests

steveeJ (approved):
Thank you @edwardgeorge for the tests and the bugfix!
The PR looks good to me. I'll leave it up to you if you want to use https://crates.io/crates/test-case to conflate these similar cases into a single test function declaration.
If you choose to do so, note that test-case doesn't support `Result` return types, i.e. it'll silently ignore errors and state that the test has passed. Which means you'd have to use `unwrap` instead of `?`.
If you choose not to do so, let me know and I'll happily merge it :slightly_smiling_face:

edwardgeorge:
@steveeJ updated with consolidated, parametrized tests

PR #180, opened 2020-10-05
client: add method to evaluate authorization status

steveeJ:
This method can be used to evaluate if a client instance is authorized
to at least call the `v2` endpoint on the remote registry.


lucab (approved):

steveeJ:
Thanks for your quick review @lucab :hugs:

PR #182, opened 2020-10-07
cargo: update release metadata

lucab:
This updates the metadata for cargo-release, in order to work
with latest release.
Notably, tags now have a `v` prefix-separator by default (see
https://github.com/sunng87/cargo-release#156).


steveeJ (approved):
Thanks!

PR #187, opened 2020-10-13
Cargo: downgrade async-stream for compatibility with rust 1.43

steveeJ:
I didn't manage to change this dependency from the consumer of dkregistry so I'm resorting to downgrading it here.

Includes #190.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ rust:
   - stable
   - beta
   - nightly
+  - 1.43.1
 cache: cargo
 matrix:
   allow_failures:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ tar = "0.4"
 tokio = "0.2"
 reqwest = { version = "0.10", default-features = false, features = ["json"] }
 sha2 = "^0.9.0"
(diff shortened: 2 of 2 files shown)


lucab:
I doubt this is really tenable because 1.43 is quite ancient at this point, but if you specifically need this because the RHEL toolset is outdated then there should be a Travis job to make sure we don't accidentally break your required MSRV with other dependencies.

steveeJ:
Thanks @lucab. I understand that 1.43 is ancient and I hope this constraint won't last for too long and doesn't create too much friction along the way.

I added 1.43.1 to the version matrix for travis in https://github.com/camallo/dkregistry-rs/pull/190

lucab:
@steveeJ the Travis change looks nice, thanks. I think you can throw that directly into this PR, so that if it is still green we can merge and be sure that everything else is ok too.

steveeJ:
Thanks @lucab. I rebased on top of #190. I think GitHub will automatically close the other one when we merge this. Let's try :slightly_smiling_face:

lucab:
LGTM, and the Travis `1.43.1` run is indeed green.

lucab (approved):

PR #198, opened 2021-02-18
Update to tokio 1.0 and reqwest 0.11

daikimiura:
All blockers seem to be gone, so tokio can be updated!
ref: https://github.com/camallo/dkregistry-rs/pull/188

@steveeJ

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.43.1
+  - 1.45.0
 cache: cargo
 matrix:
   allow_failures:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,8 @@ serde_ignored = "0.1"
 strum = "0.20"
 strum_macros = "0.20"
 tar = "0.4"
-tokio = "0.2"
-reqwest = { version = "0.10", default-features = false, features = ["json"] }
+tokio = "1.0"
+reqwest = { version = "0.11", default-features = false, features = ["json"] }
 sha2 = "^0.9.0"
 async-stream = "0.2"
 thiserror = "1.0.19"
@@ -49,7 +49,7 @@ env_logger = "0.8"
 mockito = "0.29"
 spectral = "0.6"
 test-case = "1.0.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "1.0", features = ["full"] }
 
 [features]
 default = ["reqwest-default-tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,8 @@ env_logger = "0.8"
 mockito = "0.29"
 spectral = "0.6"
 test-case = "1.0.0"
-tokio = { version = "1.0", features = ["full"] }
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.43.1
+  - 1.45.0
```
daikimiura, line 6: The minimum supported version of `tokio` is 1.45.
https://github.com/tokio-rs/tokio#supported-rust-versions

Cargo.toml
```diff
@@ -49,7 +49,7 @@ env_logger = "0.8"
 mockito = "0.29"
 spectral = "0.6"
 test-case = "1.0.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "1.0", features = ["full"] }
```
steveeJ, line 52: Is there maybe an equivalent of `"macros"`? I suspect we don't need the full featureset in this crate.

steveeJ (commented):
Minor comment on the tokio enabled features. Otherwise this looks fine.

/cc @sdodson when cincinnati will update its dkregistry dependency to a commit that includes this PR you'll probably have to bump the tokio dependency there as well.

sdodson:
@vrutkovs see above

Cargo.toml
MikailBag: AFAIK macros are still gated on the `macros` feature.

Cargo.toml
```diff
@@ -49,7 +49,8 @@ env_logger = "0.8"
 mockito = "0.29"
 spectral = "0.6"
 test-case = "1.0.0"
-tokio = { version = "0.2", features = ["macros"] }
+tokio = { version = "1.0", features = ["macros", "rt-multi-thread"] }
```
daikimiura, line 52: The `rt-multi-thread` feature is required because `tokio` has been changed to always use the multi-threaded runtime, which is only available with the `rt-multi-thread` feature.
For more information, please see the `Runtime and Builder refactor` part of [tokio's official blog (Announcing Tokio 0.3 and the path to 1.0) ](https://tokio.rs/blog/2020-10-tokio-0-3).

MikailBag: > tokio has changed to always use the multi-threaded runtime

As far as I understand, tokio still has current-thread runtime: https://docs.rs/tokio/1.2.0/tokio/runtime/struct.Builder.html#method.new_current_thread

daikimiura: As you mentioned, we can use `tokio::runtime::Builder::new_multi_thread()` to use current-thread runtime, but `tokio::runtime::Runtime::new()` will always use multi-thread runtime.
https://docs.rs/tokio/1.2.0/tokio/runtime/struct.Runtime.html#method.new

Currently `dkregistry-rs` uses `Runtime` instead of `Builder`, so unless there is a strong reason to use a current-thread runtime, I think it's better to use the multi-thread runtime without changing the existing code.

MikailBag: Ah, sorry. 

I haven't noticed that this is `dev-dependencies` change.

steveeJ (approved):
LGTM, thanks!

daikimiura:
@steveeJ 
Could you please merge this PR?

PR #201, opened 2021-03-12
Send HEAD, not GET, when checking manifest existence

tofay:
I've updated the `has_manifest` function to send a HEAD request, as the response body isn't used.

### DIFF
--- a/src/v2/manifest/mod.rs
+++ b/src/v2/manifest/mod.rs
@@ -164,7 +164,7 @@ impl Client {
         trace!("HEAD {:?}", url);
 
         let r = self
-            .build_reqwest(Method::GET, url.clone())
+            .build_reqwest(Method::HEAD, url.clone())
             .headers(accept_headers)
             .send()
             .await


vrutkovs (approved):

vrutkovs:
Seems to be allowed by https://docs.docker.com/registry/spec/api/

/cc @lucab to confirm

tofay:
> 
> 
> Seems to be allowed by https://docs.docker.com/registry/spec/api/
> 
> /cc @lucab to confirm

Yep, this PR is partly motivated by the fact that docker hub doesn't include HEAD requests in rate limit calculations: https://docs.docker.com/docker-hub/download-rate-limit/

lucab:
I'm fine to merge this as it seems spec compliant.

However please note that this initially a `HEAD` but it got turned into a `GET` in https://github.com/camallo/dkregistry-rs/pull/91, not sure if on purpose.
As such, it may (or may not) regress on Satellite repos, depending on how close to specs that implementation is at this point (they used to have many quirks, and this may or may not have been one of those).

PR #203, opened 2021-03-28
allow users to specify a prebuilt reqwest client

tofay:
if specified this will be used instead of dkregistry creating one.

This makes it easier to use non-standard authentication mechanisms:
- authenticating with Artifactory registries by setting a custom header
- setting the `Authentication` header when interacting with Azure Container Registries.

(In the latter case users login using `az` tool which writes an oauth2 token to the identitytoken field of the relevant auth section of  `$HOME/.docker/config.json`. When communicating with the registry that token can just be provided by adding the header `Authentication: Bearer <token>` )

### DIFF
--- a/src/v2/config.rs
+++ b/src/v2/config.rs
@@ -9,6 +9,7 @@ pub struct Config {
     username: Option<String>,
     password: Option<String>,
     accept_invalid_certs: bool,
+    client: Option<reqwest::Client>,
 }
 
 impl Config {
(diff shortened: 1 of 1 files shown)


tofay:
In both my use cases I want to build a reqwest client that adds a specific header, so maybe I should be allowing users to specify additional headers instead, to avoid putting `reqwest::Client` into `dkregistry`'s public API? (I think the use of reqwest is an implementation detail right now)

vrutkovs:
>maybe I should be allowing users to specify additional headers instead

Yeah, I think that would be more flexible and avoid possible API issues. IIUC https://github.com/camallo/dkregistry-rs/pull/212 is implementing that

PR #204, opened 2021-03-30
Support one-piece registry dns names

MikailBag:


MikailBag:
CI failure seems to be unrelated to proposed changes.

vrutkovs:
Breaks a test case for `library/busybox`:
```
 thread 'valid_references' panicked at '
	library/busybox:
	expected: <"library/busybox">
	 but was: <"busybox">
```

PR #211, opened 2021-11-29
make Www-Authentication header parsing case-insensitive

### LINKED ISSUE #210: Www-Authenticate header content is handled case-sensitively
Hi, I stumbled upon an issue in your library while using it with our private image registry based on Sonatype Nexus (via [OpenShift's cincinnati](https://github.com/openshift/cincinnati)).

When trying to connect and authenticate to our registry, the request fails with following error:

```text
[2021-11-19T12:33:31Z ERROR graph_builder::graph] failed to fetch all release metadata
[2021-11-19T12:33:31Z ERROR graph_builder::graph] www-authenticate header parse error
[2021-11-19T12:33:31Z ERROR graph_builder::graph] 'method' field missing
```

...where the last two lines originate from [your library](https://github.com/camallo/dkregistry-rs/blob/854d0da53bef5dd85b5e901123e85d43af97c74e/src/v2/auth.rs#L134).

Taking a deeper look at this error, I found out that our registry returns the Www-Authenticate header with the method in uppercase letters: `BASIC realm="Sonatype Nexus Repository Manager"`, but your [regex in src/v2/auth.rs](https://github.com/camallo/dkregistry-rs/blob/854d0da53bef5dd85b5e901123e85d43af97c74e/src/v2/auth.rs#L99-L110) expects the method to have its first letter uppercase with the rest in lower case. Because of this, the parsing of our registries' header fails, resulting in the above error message.

Looking at [RFC7235 Section 2.1](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1), it is noted that method and keys should be handled as case-insensitive tokens:

> It uses a **case-insensitive token as a means to identify the authentication scheme** [...]. **Authentication parameters are name=value pairs, where the name token is matched case-insensitively** [...].

I suggest changing the Www-Authentication header parsing to be case-insensitive to better conform to the HTTP specification.

I have already created a fix for this issue and will open a new pull request shortly that I kindly ask you to review.

Thanks!

ChristianLoewel:
Closes #210 

I made the following changes to make the header parsing case-insensitive:

- adapted the Regex to match method and key case-insensitively. This also required some changes in the positioning of whitespace character classes (\s) to not break the group matching.
- convert the matched method to lowercase
- rename lowercase method to PascalCase when deserializing using serde annotation
- convert all matched keys to lowercase

I also extended the relevant tests to include some variations in upper/lowercase methods and keys.

### DIFF
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -141,9 +141,7 @@ impl WwwAuthenticateHeaderContent {
                 .iter()
                 .filter_map(|capture| {
                     match (
-                        capture
-                            .name("key")
-                            .map(|n| n.as_str().to_lowercase()),
+                        capture.name("key").map(|n| n.as_str().to_lowercase()),
                         capture.name("value").map(|n| n.as_str().to_string()),
                     ) {
                         (Some(key), Some(value)) => Some(format!(
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -342,8 +342,8 @@ mod tests {
                 r#"Bearer REALM="{}",SERVICE="{}",SCOPE="{}""#,
                 realm, service, scope
             )).unwrap()
-        ] {
-            let content = WwwAuthenticateHeaderContent::from_www_authentication_header(header_value)?;
+        ].iter() {
+            let content = WwwAuthenticateHeaderContent::from_www_authentication_header(header_value.to_owned())?;
 
         assert_eq!(
             WwwAuthenticateHeaderContent::Bearer(WwwAuthenticateHeaderContentBearer {
@@ -376,9 +376,9 @@ mod tests {
             HeaderValue::from_str(&format!(r#"BASIC realm="{}""#, realm)).unwrap(),
             HeaderValue::from_str(&format!(r#"Basic Realm="{}""#, realm)).unwrap(),
             HeaderValue::from_str(&format!(r#"Basic REALM="{}""#, realm)).unwrap()
-        ] {
+        ].iter() {
             let content =
-                WwwAuthenticateHeaderContent::from_www_authentication_header(header_value)?;
+                WwwAuthenticateHeaderContent::from_www_authentication_header(header_value.to_owned())?;
 
             assert_eq!(
                 WwwAuthenticateHeaderContent::Basic(WwwAuthenticateHeaderContentBasic {
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -91,16 +91,17 @@ pub struct BasicAuth {
 
 /// Structured representation for the content of the authentication response header.
 #[derive(Debug, PartialEq, Eq, Deserialize)]
+#[serde(rename_all(deserialize = "lowercase"))]
 pub(crate) enum WwwAuthenticateHeaderContent {
     Bearer(WwwAuthenticateHeaderContentBearer),
     Basic(WwwAuthenticateHeaderContentBasic),
 }
 
 const REGEX: &str = r#"(?x)\s*
-((?P<method>[A-Z][a-z]+)\s*)?
+((?P<method>[A-Za-z]+)\s)?
+\s*
 (
-    \s*
-        (?P<key>[a-z]+)
+        (?P<key>[A-Za-z]+)
     \s*
         =
     \s*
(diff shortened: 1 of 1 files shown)


vrutkovs:
LGTM, @lucab would you like to take another look?

lucab:
LGTM, though I'm not actively maintaining nor consuming this repository anymore.

vrutkovs:
Sorry, have to close & reopen again, as Travis was replaced with Github Actions

vrutkovs:
LGTM, however `bearer_realm_parses_correctly` needs a slight update

src/v2/auth.rs
```diff
@@ -319,12 +321,29 @@ mod tests {
...
+            )).unwrap(),
+            HeaderValue::from_str(&format!(
+                r#"Bearer Realm="{}",Service="{}",Scope="{}""#,
+                realm, service, scope
+            )).unwrap(),
+            HeaderValue::from_str(&format!(
+                r#"Bearer REALM="{}",SERVICE="{}",SCOPE="{}""#,
+                realm, service, scope
+            )).unwrap()
+        ] {
```
vrutkovs, line 345:
```suggestion
].iter() {
```
To support Rust 1.46

src/v2/auth.rs
```diff
@@ -319,12 +321,29 @@ mod tests {
...
+            let content = WwwAuthenticateHeaderContent::from_www_authentication_header(header_value)?;
```
vrutkovs, line 346:
```suggestion
            let content = WwwAuthenticateHeaderContent::from_www_authentication_header(header_value.to_owned())?;
```
for Rust 1.46 support

src/v2/auth.rs
```diff
@@ -350,16 +370,23 @@ mod tests {
...
-            }),
-            content
-        );
+        for header_value in [
+            HeaderValue::from_str(&format!(r#"Basic realm="{}""#, realm)).unwrap(),
+            HeaderValue::from_str(&format!(r#"basic realm="{}""#, realm)).unwrap(),
+            HeaderValue::from_str(&format!(r#"BASIC realm="{}""#, realm)).unwrap(),
+            HeaderValue::from_str(&format!(r#"Basic Realm="{}""#, realm)).unwrap(),
+            HeaderValue::from_str(&format!(r#"Basic REALM="{}""#, realm)).unwrap()
+        ] {
```
vrutkovs, line 379: Same here - `.iter()` and `.to_owned()` later

ChristianLoewel:
Thanks, @vrutkovs! I updated the PR accordingly.

src/v2/auth.rs
```diff
@@ -133,14 +134,15 @@ impl WwwAuthenticateHeaderContent {
...
+            .to_lowercase()
             .to_string();
 
         let serialized_content = {
             let serialized_captures = captures
                 .iter()
                 .filter_map(|capture| {
                     match (
-                        capture.name("key").map(|n| n.as_str().to_string()),
+                        capture.name("key").map(|n| n.as_str().to_lowercase().to_string()),
```
vrutkovs, line 145:
```suggestion
                        capture
                            .name("key")
                            .map(|n| n.as_str().to_lowercase().to_string()),
```
See `cargo fmt` from lints check

src/v2/auth.rs
```diff
@@ -133,14 +134,17 @@ impl WwwAuthenticateHeaderContent {
             .name("method")
             .ok_or(WwwHeaderParseError::FieldMethodMissing)?
             .as_str()
+            .to_lowercase()
             .to_string();
```
vrutkovs, line 138: No longer needed:
```
 Error:    --> src/v2/auth.rs:138:13
    |
138 |             .to_string();
    |             ^^^^^^^^^^^^ help: remove this
    |
    = note: `-D clippy::redundant-clone` implied by `-D warnings`
```

src/v2/auth.rs
```diff
@@ -133,14 +134,17 @@ impl WwwAuthenticateHeaderContent {
...
 
         let serialized_content = {
             let serialized_captures = captures
                 .iter()
                 .filter_map(|capture| {
                     match (
-                        capture.name("key").map(|n| n.as_str().to_string()),
+                        capture
+                            .name("key")
+                            .map(|n| n.as_str().to_lowercase().to_string()),
```
vrutkovs, line 147:
```suggestion
                            .map(|n| n.as_str().to_lowercase()),
```
`to_lowercase` returns String

vrutkovs (changes requested):

ChristianLoewel:
yet another formatting change. Sorry for this back and forth, but now it should be okay.

vrutkovs:
Looks great now, thanks!

PR #212, opened 2021-12-01
Allow passing custom set of accepted types with optional weight

vrutkovs:
Some registries don't parse weight parameter correctly, some require custom parameter weights.

This commits allows passing custom set of accepted types (with optional weight) and moves GCR workaround from v2/manifest/mod.rs to config.rs.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=2010497


vrutkovs:
cc @PratikMahajan @wking

PratikMahajan (approved):

PR #216, opened 2021-12-01
github actions: move test-net-private tests to a separate suite

vrutkovs:
test-net-private tests may be unstable, so we should not block on these

### DIFF
--- a/.github/workflows/minimal.yml
+++ b/.github/workflows/minimal.yml
@@ -67,6 +67,29 @@ jobs:
           command: test
           args: --features test-mock
 
+  test-net-private:
+    name: Test Suite (network-enabled tests)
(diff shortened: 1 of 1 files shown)


.github/workflows/minimal.yml
```diff
@@ -67,6 +67,29 @@ jobs:
           command: test
           args: --features test-mock
 
+  test-net-private:
+    name: Test Suite (network-enabled tests)
```
lucab, line 71: I think this actually needs porting some service account tokens that Travis has.

vrutkovs: I did add the token as a secret and this test passed a few times - but now Quay throws 429. Probably my sprint of PRs has overloaded it

PR #224, opened 2022-02-07
Fetch blob as stream

anti-social:
Here we need this functionality: https://github.com/tailhook/vagga/pull/572

### DIFF
--- a/src/v2/content_digest.rs
+++ b/src/v2/content_digest.rs
@@ -32,7 +32,7 @@ pub enum ContentDigestError {
 /// ContentDigest stores a digest and its DigestAlgorithm
 #[derive(Clone, Debug)]
 pub struct ContentDigest {
-    expected_digest: String,
+    digest: String,
     algorithm: DigestAlgorithm,
 }
 
@@ -51,21 +51,21 @@ impl ContentDigest {
 
         let algorithm = std::str::FromStr::from_str(digest_split[0])?;
         Ok(ContentDigest {
-            expected_digest: digest.to_string(),
+            digest: digest.to_string(),
             algorithm,
         })
     }
 
-    pub fn hash(&mut self, input: &[u8]) {
+    pub fn update(&mut self, input: &[u8]) {
         self.algorithm.update(input)
     }
 
     pub fn verify(self) -> std::result::Result<(), ContentDigestError> {
         let digest = self.algorithm.digest();
-        if digest != self.expected_digest {
+        if digest != self.digest {
             return Err(ContentDigestError::Verify {
-                expected: self.expected_digest.clone(),
-                got: digest.clone(),
+                expected: self.digest,
+                got: digest,
             });
         }
         Ok(())
@@ -74,7 +74,7 @@ impl ContentDigest {
 
 impl std::fmt::Display for ContentDigest {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}:{}", self.algorithm, "self.digest")
+        write!(f, "{}:{}", self.algorithm, self.digest)
     }
 }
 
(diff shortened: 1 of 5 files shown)


src/v2/content_digest.rs
```diff
@@ -1,100 +1,122 @@
...
-        got: ContentDigest,
+        expected: String,
+        got: String,
     },
 }
 
+/// ContentDigest stores a digest and its DigestAlgorithm
+#[derive(Clone, Debug)]
+pub struct ContentDigest {
+    expected_digest: String,
```
vrutkovs, line 42: Not sure if it needs to be renamed, its unclear what "expected" means here

src/v2/content_digest.rs
```diff
@@ -1,100 +1,122 @@
...
-
-        trace!("content verification succeeded for '{}'", &layer_digest);
         Ok(())
     }
 }
 
 impl std::fmt::Display for ContentDigest {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}:{}", self.algorithm, self.digest)
+        write!(f, "{}:{}", self.algorithm, "self.digest")
```
vrutkovs, line 84: I don't think quotes are necessary - typo?

src/v2/content_digest.rs
```diff
@@ -117,25 +139,39 @@ mod tests {
     }
 
     #[test]
-    fn try_verify_succeeds_with_same_content() -> Fallible<()> {
+    fn verify_content_ok() -> Fallible<()> {
```
vrutkovs, line 142:
```suggestion
    fn verify_succeeds_with_same_content() -> Fallible<()> {
```
so that we keep in line with `verify_fails_with_different_content`

vrutkovs (commented):
Thank you for your PR! Some notes below

src/v2/content_digest.rs
```diff
@@ -1,100 +1,122 @@
...
         })
     }
 
-    /// try_verify hashes the input slice and compares it with the digest stored in this instance
-    ///
-    /// Success depends on the result of the comparison
-    pub fn try_verify(&self, input: &[u8]) -> std::result::Result<(), ContentDigestError> {
-        let hash = self.algorithm.hash(input);
-        let layer_digest = Self::try_new(hash)?;
+    pub fn hash(&mut self, input: &[u8]) {
```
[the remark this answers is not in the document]
anti-social: I would rename it to `update` - it is well known name when hashing data.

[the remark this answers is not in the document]
vrutkovs: good idea

vrutkovs:
Looks good! Lets add some unit tests - i.e. https://github.com/vrutkovs/dkregistry-rs/commit/325e87b2f8663a2897c6735cd18512cba2c09e18

anti-social:
> Looks good! Lets add some unit tests - i.e. [vrutkovs@325e87b](https://github.com/vrutkovs/dkregistry-rs/commit/325e87b2f8663a2897c6735cd18512cba2c09e18)

Oh, thanks. Yesterday also wanted to add tests but don't find out why network tests failed.

vrutkovs:
Network tests are passing, but in Github Actions we get 429 from quay. Investigating this in #222

anti-social:
@vrutkovs We want to display a progress bar while downloading a layer. At the moment we cannot do that as there is no api to know layer's size. I suggest to slightly change a fetching blob api like following:

```rust
let blob_resp = dkclient.get_blob_response().await?;
let blob_size = blob_resp.size();
let blob_stream = blob_resp.stream();
```

anti-social:
Implemented getting blob size when using stream api.

vrutkovs:
This looks great, thank you! Lets merge this now and adjust later if necessary

PR #232, opened 2022-07-07
add support for ManifestList images

PratikMahajan:
ManifestList images also known as fat-manifests are
collections of manifests for different arch/os.
this pr adds support for ManifestList images.

As manifest list images only contain digests of the
images contained in the manifest, the `layers_digests`
function returns the digest of all the images
contained in the ManifestList instead of individual
layers of the manifests.
The layers of a specific image from manifest list can
be obtained using the digest of the image from the
manifest list and getting it's manifest and manifestref
(get_manifest_and_ref()) and using this manifest of
the individual image to get the layers.

### DIFF
--- a/src/v2/config.rs
+++ b/src/v2/config.rs
@@ -112,8 +112,7 @@ impl Config {
                     // https://tools.ietf.org/html/rfc7231#section-5.3.2
                     (MediaTypes::ManifestV2S2, Some(0.5)),
                     (MediaTypes::ManifestV2S1Signed, Some(0.4)),
-                    // TODO(steveeJ): uncomment this when all the Manifest methods work for it
-                    // mediatypes::MediaTypes::ManifestList,
+                    (MediaTypes::ManifestList, Some(0.5)),
                 ],
                 // GCR incorrectly parses `q` parameters, so we use special Accept for it.
                 // Bug: https://issuetracker.google.com/issues/159827510.
@@ -121,6 +120,7 @@ impl Config {
                 true => vec![
                     (MediaTypes::ManifestV2S2, None),
                     (MediaTypes::ManifestV2S1Signed, None),
+                    (MediaTypes::ManifestList, None),
                 ],
             },
         };
--- a/src/v2/manifest/manifest_schema2.rs
+++ b/src/v2/manifest/manifest_schema2.rs
@@ -141,3 +141,27 @@ impl ManifestSchema2 {
         self.config_blob.architecture.to_owned()
     }
 }
+
+impl ManifestObj {
+    /// Get the architecture of the manifest object
+    pub fn architecture(&self) -> String {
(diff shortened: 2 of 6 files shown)


src/v2/config.rs
```diff
@@ -112,15 +112,15 @@ impl Config {
...
-                    // mediatypes::MediaTypes::ManifestList,
+                    (MediaTypes::ManifestList, Some(0.5)),
                 ],
                 // GCR incorrectly parses `q` parameters, so we use special Accept for it.
                 // Bug: https://issuetracker.google.com/issues/159827510.
                 // TODO: when bug is fixed, this workaround should be removed.
                 true => vec![
                     (MediaTypes::ManifestV2S2, None),
                     (MediaTypes::ManifestV2S1Signed, None),
+                    (MediaTypes::ManifestList, None),
```
wking, line 123: worth working in support for [`application/vnd.oci.image.index.v1+json`][1] at the same time, or different enough that we should defer that to later work?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.2/media-types.md#applicationvndociimageindexv1json

src/v2/manifest/mod.rs
```diff
@@ -295,8 +295,20 @@ pub enum ManifestError {
 
 impl Manifest {
     /// List digests of all layers referenced by this manifest, if available.
+    /// For ManifestList, returns the digest of all the manifest list images.
```
wking, line 298: nit: `digest` -> `digests`

src/v2/config.rs
PratikMahajan: We currently don't support OCI media types in dkregistry. https://github.com/camallo/dkregistry-rs/blob/master/src/mediatypes.rs

IMO we should be adding the support for OCI media types separately, as we'll be only supporting a single OCI media type if we end up adding [application/vnd.oci.image.index.v1+json](https://github.com/opencontainers/image-spec/blob/v1.0.2/media-types.md#applicationvndociimageindexv1json) to this PR

src/v2/manifest/manifest_schema2.rs
```diff
@@ -141,3 +141,32 @@ impl ManifestSchema2 {
...
+impl ManifestList {
+    /// Get architecture of all the manifests
+    pub fn architectures(&self) -> Vec<String> {
+        self.manifests.iter().map(|mo|{
+            mo.architecture()
+        }).collect()
+    }
+
+    /// Get the digest for all the manifest images in the ManifestList
+    pub fn get_digest(&self) -> Vec<String> {
```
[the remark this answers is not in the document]
PratikMahajan: renamed

src/v2/manifest/mod.rs
PratikMahajan: changed

PR #236, opened 2022-08-01
Modified Tag Pagination to get all query parameters from Link Header.

kobutton:
This library does not work for listing tags with some registries such as JFrog's Artifactory, as the Artifactory Docker Registry does not use the "next_page" url query parameter. 
This Pull Request solves that issue by extracting all the query parameters returned in the Link Header provided by the registry server and uses them to fetch the next page of tags. 
This ensures a better alignment with the spec for tag pagination as defined by the docker API v2 Spec [Here](https://docs.docker.com/registry/spec/api/#pagination-1), while also maintaining compatibility with servers that use the "next_page" url query parameter such as quay.


PratikMahajan (approved):
/lgtm

PR #239, opened 2022-11-08
Add Config::add_root_certificate for trusting custom ca

iamjpotts:
Support connecting to registries with self-signed certificates and certificates signed by CAs that are not in the client system trust store.

Simply passes thru the certificate along to a `reqwest` builder method by the same name.

Adds two tests:

1. Verify the client fails with an SSL error when the CA is not provided
2. Verify the client connects without an SSL error when the CA is given

Both tests are a part of the `test-mock` feature.


.gitignore
```diff
@@ -3,3 +3,4 @@ Cargo.lock
 .envrc
 default.nix
 shell.nix
+.idea
```
PratikMahajan, line 6: this need not be in the commit. Depends on dev workflow, ide.
Can we worked around by using gobal .gitignore on dev system

PratikMahajan (changes requested):
PR needs rebase

.gitignore
iamjpotts: Removed.

iamjpotts:
> PR needs rebase

Surprised, what did you notice that made it look like it needed a rebase? It was rebased about 5 months ago.

It should be current with the default branch now in any case.

iamjpotts:
When this PR was originally authored 2 years ago, `hyper 1.x` had not been released.

Do you want me to upgrade this PR to hyper 1.x, or leave it on version 0.14?

iamjpotts:
> can you upgrade to hyper 1.x and it'll be good to go!

Done.

PratikMahajan (approved):
/lgtm

PratikMahajan:
looks like we need to move to [v0.14.28](https://github.com/hyperium/hyper/releases/tag/v0.14.28)
sorry about that

iamjpotts:
* Back to hyper 0.14
* Fixed `cargo fmt` issues

PR #244, opened 2023-01-27
Add support for OCI Image Index

### LINKED ISSUE #243: 404 when retrieving manifest
I have noticed an increasing number of Docker Hub repositories are returning 404 when retrieving manifests. An example of such repositories is `koenkk/zigbee2mqtt:latest`.

Any idea what might be happening?

luizribeiro:
To be honest, I don’t know enough about how Docker image manifests work, but from looking at the documentation it seems like OCI Image Index follows exactly the same format as Manifest Lists. So this makes us use that format for whenever we encounter the OCI Image Index MIME type.

This fixes #243 for me.

### DIFF
--- a/src/mediatypes.rs
+++ b/src/mediatypes.rs
@@ -32,6 +32,10 @@ pub enum MediaTypes {
     #[strum(serialize = "application/vnd.docker.container.image.v1+json")]
     #[strum(props(Sub = "vnd.docker.container.image.v1+json"))]
     ContainerConfigV1,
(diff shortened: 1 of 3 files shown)


PratikMahajan (approved):

luizribeiro:
Thank you for reviving this @chantra, I appreciate it :)

luizribeiro:
Closing due to lack of activity. Happy to reopen if this is still needed!

PR #253, opened 2023-05-29
Make sure compiling without tls support is possible

jcfj:
Without this PR, compiling the current master with `no-default-features` fails with
```none
error[E0599]: no method named `danger_accept_invalid_certs` found for struct `ClientBuilder` in the current scope
   --> /tmp/fork-dkregistry-rs/src/v2/config.rs:104:14
    |
103 |           let client = reqwest::ClientBuilder::new()
    |  ______________________-
104 | |             .danger_accept_invalid_certs(self.accept_invalid_certs)
    | |             -^^^^^^^^^^^^^^^^^^^^^^^^^^^ method not found in `ClientBuilder`
    | |_____________|
    | 

For more information about this error, try `rustc --explain E0599`.
```

I've opted for making the Client builder functions that make no sense without tls (`insecure_registry` and `accept_invalid_certs`) configuration dependent. Whether that is the right thing to do I don't know, but it might allow a user of the crate to notice specifying the wrong features at compile time.

(I've also bumped the MSVR in the workflow so I can see whether this builds successfully. The CI is still angry, but that looks unrelated.)


vrutkovs:
cc @PratikMahajan

jcfj:
Hmm, I thought about this a bit more. Maybe it'd be better to just allow passing a reqwest client instance like [so](https://github.com/camallo/dkregistry-rs/commit/3c8c13817ec0eef46a2bbf59437b1e97dc13214c). Reqwest can [deal](https://docs.rs/reqwest/0.11.18/src/reqwest/async_impl/client.rs.html#1220-1228) with all the cfg stuff, and it'd add some flexibility.

PR #254, opened 2023-06-01
Ci fixes

jeckersb:
This should fix (most of) the current issues blocking successful CI runs.  There is still an issue with the network-enabled tests failing due to an HTTP 429 error from Quay, presumably from rate-limiting github actions.


vrutkovs:
Thank you, merged!

PR #260, opened 2024-02-08
API error: improve error message

chantra:
Previously, a generic error would be returned:
```
Error: unexpected HTTP status 404 Not Found
```

This does not convey much information as to what happens. On the current version of dkregistry, one would hit this when trying to access s390x/ubuntu on registry-1.docker.io.

This change specializes the error to surface more information as to why the error is happening as described in the OCI Distribution Specification: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#error-codes

In this specific case, the error would look like:
```
Error: Api Error: ((MANIFEST_UNKNOWN), message: OCI manifest found, but accept header does not support OCI manifests)

Caused by:
    ((MANIFEST_UNKNOWN), message: OCI manifest found, but accept header does not support OCI manifests)
```

which clearly highlight that dkregistry does not support OCI manifest.


chantra:
I need a minor change and will add testing.

chantra:
added test against fixtures and a test-net test.

chantra:
@PratikMahajan any possibilities you look into this PR?

PR #261, opened 2024-02-09
Add support for OCI Image Manifest

chantra:
Image like s390x/ubuntu uses OCI Image Manifest, which per
https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md
is very similar to vnd.docker.distribution.manifest.v2.

Add a test to confirm we can deserialize an OCI Image Manifest as a ManifestSchema2Spec.

Also add a test to confirm that we can now fetch the manifest from docker.io.
Previously it would fail with MANIFEST_UNKNOWN.

This is built on top of #260 . I stacked it because it would otherwise have merge conflicts due to the tests.


chantra:
As a follow up, I am happy to revive #244 if @luizribeiro can't get to it.

chantra:
@PratikMahajan if you could look into this one too, that would be great.

PratikMahajan (approved):

Issue #262, opened 2024-02-28
Authenticating against Azure Container Registry (ACR) fails


ThorstenHans:
The current implementation does not work with Azure Container Registry (ACR). As the REST specification outlines, Microsoft returns the access token using the `access_token` field instead of a field called `token`.

Ref: https://learn.microsoft.com/en-us/rest/api/containerregistry/access-tokens/get-from-login?view=rest-containerregistry-2019-08-15&tabs=HTTP#get-access-token-with-username-and-password

Allowing deserialization from both fields would allow authentication with Azure Container Registry as well.

PR #263, opened 2024-02-28
Add support for authenticating against Azure Container Registry

ThorstenHans:
As described in #262 Azure Container Registry returns the access token using the `access_token` field instead of the `token` field. With this commit, the token could be fetched from both fields

### DIFF
--- a/src/v2/auth.rs
+++ b/src/v2/auth.rs
@@ -32,8 +32,9 @@ pub struct BearerAuth {
     refresh_token: Option<String>,
 }
 
+/// Used to support different response schemas of Bearer HTTP Authentication
 #[derive(Debug, Clone, Default, Deserialize)]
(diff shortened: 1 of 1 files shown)


ThorstenHans:
Upon further testing I discovered that Docker Hub returns both fields `token` and `access_token` in the response. I know that the name of the added struct (`AzureBearerAuth`) is not optimal. But using an intermediate struct allowed me to successfully authenticate against both Docker Hub and ACR.

vrutkovs:
cc @PratikMahajan

ThorstenHans:
Renamed the struct from `AzureBearerAuth` to `MultiTokenBearerAuth` which is a bit more meaningful

PratikMahajan (approved):

